PR #85, opened 2017-02-10
Fixes in examples (PyOpenGL and PyQT5)

Aljenci:
PyOpenGL:
The call to glutCreateWindow want the argument to be of type byte no str.
Error: ctypes.ArgumentError: argument 1: <class 'TypeError'>: wrong type

Changed glutCreateWindow('') to glutCreateWindow(b'').

PyQT5 (Raytracer example):
Crash using the iface. Changed to use only the prog
prog, iface = GL.NewProgram([vert, frag])
TypeError: 'ModernGL.Program' object is not iterable

------------------------
Also added a name to the windows to have the example name in the PyOpenGL examples.


cprogrammer1994:
Thank you

PR #140, opened 2017-10-09
[Context] Add support for float32 texture internal formats

toaarnio:
Previously, float32 textures were normalized to [0, 1] by the Texture
constructor.


cprogrammer1994:
Thank you!

PR #150, opened 2017-11-20
Added get method to Map-like objects

fekga:
### Description

Map-like objects didn' have any `get` method for the underlying `dict` member.

### List of changes

- **added** Each Map has now a `get` method with the default value of `None`.


cprogrammer1994:
Thank you!

PR #186, opened 2018-05-08
Tesselation example

joshua-r:
### Description

Add example showing how to use tesselation.

### List of changes

- **added** an example showing how to draw a cubic Bézier curve using tesselation.
- **changed** list of contributors in the readme


joshua-r:
I hope it's fine that I'm using glfw and not the examples-module. I have been using glfw all the time and there was no example for it yet. Feel free to change that if you want the examples to be consistent.

PR #191, opened 2018-06-10
Example fixes

einarf:
### Description

Misc fixes when going through the examples directory.

### List of changes

- Ensure example windows support 4k and retina displays (viewport was incorrect)
- GLFW example did not specify core forward compat context causing issues for some setups
- Simplify README: Only require `pip install -r requirements.txt`
- Added matplotlib to requirements.txt
- A couple of `.gitignore` entries

### DIFF
--- a/examples/04_colors_and_texture.py
+++ b/examples/04_colors_and_texture.py
@@ -6,7 +6,7 @@ import os
 
 import moderngl
 import numpy as np
-from objloader import Obj
+from moderngl_obj import Obj
 from PIL import Image
 from pyrr import Matrix44
 
--- a/examples/04_colors_and_texture.py
+++ b/examples/04_colors_and_texture.py
@@ -6,7 +6,7 @@ import os
 
 import moderngl
 import numpy as np
-from moderngl_obj import Obj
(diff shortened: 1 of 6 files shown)


examples/04_colors_and_texture.py
```diff
@@ -6,7 +6,7 @@
 
 import moderngl
 import numpy as np
-from objloader import Obj
+from moderngl_obj import Obj
```
cprogrammer1994: `moderngl_obj` is deprecated and `objloader` should be used instead.

einarf: I reverted that further down when I checked the age of the packages on pipy

einarf:
Ignore the second commit 😄

einarf:
I can submit a cleaner pr without the second commit if you want, or do you auto-squash on merge? This was just things I picked up in my journey through the examples directory over time, so I derped on the obj package initially.

cprogrammer1994:
Oh, I just realized that you already reverted that change. The final diff looks great. merging

cprogrammer1994:
Thank you!

PR #195, opened 2018-06-25
Framebuffer.clear changes global clear values

einarf:
### Description

I'm not entirely sure if this is the right thing to do, but `fbo.clear()` is currently changing the global clear values. This can be very confusing in more complex programs.

### List of changes

- FrameBuffer.clear() restores the original clear color values

### DIFF
--- a/src/Framebuffer.cpp
+++ b/src/Framebuffer.cpp
@@ -376,6 +376,13 @@ PyObject * MGLFramebuffer_clear(MGLFramebuffer * self, PyObject * args) {
 
 	gl.BindFramebuffer(GL_FRAMEBUFFER, self->framebuffer_obj);
 	gl.DrawBuffers(self->draw_buffers_len, self->draw_buffers);
+
+	// Store the current clear values
+	float color_clear_value[4];
+	float depth_clear_value;
+	gl.GetFloatv(GL_COLOR_CLEAR_VALUE, color_clear_value);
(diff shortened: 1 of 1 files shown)


einarf:
Again, not entirely sure if this is a good idea. It might make sense for people who use a more mixed PyOpenGL / ModernGL setup. It could also be argued one should use `ctx.clear(r, g, b, a)` and never mix with `glClear`.

Feel free to close if you think this is a bad idea.

cprogrammer1994:
Hi!

I tried to avoid using glGet in functions used in the main loop of a program.
In my experience they can cause performance drop on some hardware, even if their implementation may not involve the GPU at all.

In the future `fbo.clear` will use `glClearBuffer` but these changes are experimental.
https://github.com/cprogrammer1994/ModernGL/blob/532ce157c0a2572541e642bc4c605f0e6f7fce6c/moderngl/mgl/framebuffer.cpp#L209

The experimental branch takes a lot of time to be compatible with all the tests & examples of the master.

PS:
Mixing PyOpenGL and ModernGL can be error prone with the bound framebuffer too.
The Context have a bound_framebuffer (set by `fbo.use`). Framebuffer operation that require a bind will restore this value as follows:
https://github.com/cprogrammer1994/ModernGL/blob/e480d5d37c5015e2f11daf9c5a81fe7ac1803c5a/src/Framebuffer.cpp#L547

Thank you!

einarf:
Yup. Let's close this one. Thanks for the info! Will take a peak in the experimental branch.

PR #223, opened 2018-07-28
Missing sampler parameters

einarf:
### Changes

* Bug: ``repeat_y`` was returning ``repeat_x`` in ``context.py``. Added tests for this.
* Bug: glError on sampler creation
* Added ``repeat_z``, ``min_lod``, ``max_lod`` and ``border_color``
* Added ``Context.clear_samplers``

### DIFF
--- a/tests/test_sampler.py
+++ b/tests/test_sampler.py
@@ -15,21 +15,26 @@ class TestCase(unittest.TestCase):
         sampler = self.ctx.sampler()
 
         # Default values
-        assert sampler.anisotropy == 1.0
-        assert sampler.repeat_x is True
-        assert sampler.repeat_y is True
-        assert sampler.filter == (moderngl.LINEAR, moderngl.LINEAR)
-        assert sampler.compare_func == '?'
+        self.assertEqual(sampler.anisotropy, 1.0)
+        self.assertTrue(sampler.repeat_x)
+        self.assertTrue(sampler.repeat_y)
+        self.assertTrue(sampler.repeat_z)
+        self.assertEqual(sampler.filter, (moderngl.LINEAR, moderngl.LINEAR))
+        self.assertEqual(sampler.compare_func, '?')
 
         # Change values
-        sampler.anisotropy = 2.0
-        sampler.repeat_x = False
-        sampler.repeat_y = False
+        sampler.anisotropy = self.ctx.max_anisotropy
         sampler.filter = (moderngl.NEAREST_MIPMAP_NEAREST, moderngl.NEAREST)
         sampler.compare_func = "<="
 
-        assert sampler.anisotropy == 2.0
-        assert sampler.repeat_x is False
-        assert sampler.repeat_y is False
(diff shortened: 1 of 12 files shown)


einarf:
Adding the last three parameters quickly as well.

einarf:
The only wonky part with border color is that it requires a different texture wrap having to override ``repeat_``. When the border color is set I simply override it for now.

Testing with yellow border color:
screen shot 2018-07-28 at 15 12 22

einarf:
Done for now. I have tested the new parameters.

einarf:
Also added ``Contect.clear_samplers`` unbinding samplers from all texture units. When you for example draw a GLTF scene (that uses samplers for all textures) you really have no idea what state the texture units are currently in and you have no way to efficiently clean them up.

It's probably not ideal, but MUCH more efficient than unbinding samplers for every mesh.

einarf:
Doing some more detailed testing. I see a glError triggering somewhere I have to find as well.

einarf:
What do you think about having the sampler as an optional parameter to ``Texture.use(...)``?

Currently:
```python
sampler.use(location=0)
texture.use(location=0)
```

Sampler parameter:
```python
texture.use(location=1, sampler=sampler)
```

It may give the impression that a sampler belongs to the bound texture and not the texture unit as a whole? .. or we assume people know the sampler will stick to the unit?

cprogrammer1994:
Is it common to create a few samplers and use with multiple textures?
What do you think, should samplers be created like vao's? (Vaos store a program object. When you call vao.render it will bind the program object that vao internally knows. Sampler.use could bind the texture to a texture unit)
I think it could be more convenient to have samplers as the only option to bind a texture to a texture unit (in the next major). Textures would store the texture's data only. Samplers would be responsible for handling the sampler parameters and binding the texture to its unit when requested.
This could ensure not to have invalid sampler, texture pairs bound.
This could also ensure that `build_mipmaps` cannot affect the filter parameters.
This would force the user to create similar samplers for different textures.

tests/test_sampler.py
```diff
@@ -4,32 +4,87 @@
 
 from common import get_context
 
+def checkerror(func):
```
cprogrammer1994: I think we need no checks on gl errors. The initialization step results in gl errors for some gl get calls. I don't want to clear them. I also implemented a standalone context that uses the `GL_CONTEXT_FLAG_NO_ERROR_BIT`.
Sometimes I install hooks on the opengl calls to check for gl errors (unfortunately these calls will ensure error=0).
I think tests should not depend on opengl error values since they are changed when inspected.

einarf:
I moved my question about sampler binding to #227

tests/test_sampler.py
einarf: Fair enough.

cprogrammer1994:
I will merge this. Can we make a release when this one is merged?

einarf:
Yup. Release 😄

PR #232, opened 2018-07-31
Missing program_members package in release

einarf:
### Description

The ``moderngl.program_members`` package is not included in the distributions.


einarf:
Optionally just use ``find_packages*`` to make it more future proof.

cprogrammer1994:
I reverted the `find_packages` just before making a release. :(

einarf:
It's not supposed to be easy 😃

PR #234, opened 2018-08-01
Fix a regression in Context.cpp

### LINKED ISSUE #233: Regression in 5.4.x related to off-screen render targets
**Description**

There is a regression related to off-screen render targets in 5.4.x. After using a 2-channel off-screen framebuffer, the main framebuffer starts acting like a 2-channel framebuffer as well. From then on, only the first two channels are affected by any operation, including `clear()`.

**Proof of Code**

```py
#!/usr/bin/python3 -B

import numpy as np          # pip install numpy
import moderngl as gl       # pip install moderngl
import pygame               # pip install pygame

def main():
    winsize = (4, 4)
    pygame.display.init()
    pygame.display.set_mode(winsize, pygame.OPENGL)
    ctx = gl.create_context()
    offscreen = ctx.simple_framebuffer(winsize, components=2, dtype='f1')
    offscreen.use()
    ctx.screen.use()
    ctx.clear(0.50, 0.50, 0.50, 1.0)
    pixels = ctx.screen.read(winsize, components=3, dtype='f1')
    pixels = np.frombuffer(pixels, dtype=np.uint8).reshape(4,4,3)
    print("EXPECTED: [128 128 128]")
    print("ACTUAL  :", pixels[0,0])
    assert(np.all(pixels[0,0] == [128, 128, 128]))

if __name__ == "__main__":
    main()
```

Result on 5.3.0:
```
$ ./moderngl-bug.py
EXPECTED: [128 128 128]
ACTUAL  : [128 128 128]
```

Result on 5.4.2:
```
$ ./moderngl-bug.py
EXPECTED: [128 128 128]
ACTUAL  : [128 128 0]
  File "./moderngl-bug.py", line 24, in <module>
    main()
  File "./moderngl-bug.py", line 20, in main
    assert(np.all(pixels[0,0] == [128, 128, 128]))
AssertionError
```

**Version info**

```
moderngl 5.4.2
--------------
vendor: Intel Open Source Technology Center
renderer: Mesa DRI Intel(R) HD Graphics 630 (Kaby Lake GT2) 
version: 4.5 (Core Profile) Mesa 18.0.5
python: 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609]
platform: linux
code: 450
```

toaarnio:
framebuffer->draw_buffers_len was initialized to 0 instead of 1 in
MGLContext_Initialize().

This fixes #233.

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -1372,7 +1372,7 @@ void MGLContext_Initialize(MGLContext * self) {
 
 		framebuffer->framebuffer_obj = 0;
 
-		framebuffer->draw_buffers_len = 0;
+		framebuffer->draw_buffers_len = 1;
 		framebuffer->draw_buffers = new unsigned[1];
 
 		// According to glGet docs:


cprogrammer1994:
Are you sure this affects the clear?

the Framebuffer.clear and the Framebuffer.use ignore this value.

toaarnio:
It does, but I don't know how. I found the bug with `git bisect`, so I didn't (need to) do any further analysis or debugging. It may well be that there's a deeper root cause hidden somewhere else, and this is only curing the symptoms.

PR #238, opened 2018-08-05
Properly handle flags when enter/exit scope

### LINKED ISSUE #237: Scope do not revert to previous enabled flags
**Description**

I cannot see ``old_enable_flags`` ever being assigned a value when entering the scope, so the exit code below will just disable everything.

https://github.com/cprogrammer1994/ModernGL/blob/6aff80bc14d14a8418175acd3f607eb595d2145b/src/Scope.cpp#L202-L228

**Expected behavior**

According to the docs, enable flags should be restored when the scope reverts: https://moderngl.readthedocs.io/en/stable/reference/scope.html#examples

**Reproducing Code**

Output:
```
Before scope: DEPTH_TEST 1
Before scope: BLEND 0
In scope: DEPTH_TEST 0
In scope: BLEND 1
After scope: DEPTH_TEST 0
After scope: BLEND 0
```
Code:
```py
from OpenGL import GL
import moderngl

ctx = moderngl.create_standalone_context()
ctx.enable(moderngl.DEPTH_TEST)

size = (512, 512)
fbo = ctx.framebuffer(
    ctx.texture(size, 4),
    depth_attachment=ctx.depth_texture(size),
)
scope = ctx.scope(fbo, enable_only=moderngl.BLEND)

def print_flags(label):
    print(label, 'DEPTH_TEST', GL.glGetInteger(GL.GL_DEPTH_TEST))
    print(label, 'BLEND', GL.glGetInteger(GL.GL_BLEND))

print_flags("Before scope:")

with scope:
    print_flags("In scope:")

print_flags("After scope:")
```

**Version info**

```
moderngl 5.4.2
--------------
vendor: NVIDIA Corporation
renderer: GeForce GTX 1070/PCIe/SSE2
version: 4.6.0 NVIDIA 391.01
python: 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)]
platform: win32
code: 460
```

einarf:
### Description

Fixes #237 

When entering a scope we update context with the new flags storing the old ones. Once the scope reverts we restore the old flags.

### DIFF
--- a/docs/reference/scope.rst
+++ b/docs/reference/scope.rst
@@ -46,7 +46,7 @@ Examples
 
     scope = ctx.scope(
         framebuffer=framebuffer1,
-        enable_flags=moderngl.BLEND,
+        enable_only=moderngl.BLEND,
         textures=[
             (texture1, 4),
             (texture2, 3),
--- a/moderngl/scope.py
+++ b/moderngl/scope.py
@@ -9,9 +9,9 @@ class Scope:
 
         - Set the enable flags.
         - Bind the framebuffer.
-        - Assing textures to texture locations.
-        - Assing buffers to uniform buffers.
-        - Assing buffers to shader storage buffers.
+        - Assigning textures to texture locations.
+        - Assigning buffers to uniform buffers.
+        - Assigning buffers to shader storage buffers.
 
         Responsibilities on exit:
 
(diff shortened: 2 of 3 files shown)


einarf:
Not entirely sure how important it is to keep track of the enable flags in the context itself, but at least this way it should be safe to use nested scopes.

cprogrammer1994:
It is important to have the proper value in the enable_flags. Thank you!

PR #240, opened 2018-08-11
Expose extra attributes + initialize

einarf:
1) Expose the ``extra`` attribute in docs for all types. This is a fairly important attribute to be aware of since moderngl types cannot be extended using inheritance. I personally went down an unecessary rabbit hole of making wrappers though composition before I saw the ``extra`` attribute.

2) Initialize all ``extra`` attributes on instance creation. ``program.extra`` will trigger ``AttributeError`` unless the ``extra`` attribute has previously been assigned a value. I think it's a lot more user friendly to ensure the attribute is present with a default ``None`` value. 

This default value is also advertised in most initializers, so most IDEs and code inspection tools believe the attribute is present with a default value. (They don't care about the ``TypeError()``)

https://github.com/cprogrammer1994/ModernGL/blob/a2cf2e1b20939331b25e63531f128b4f9008ab62/moderngl/program.py#L26-L34

### DIFF
--- a/docs/reference/attribute.rst
+++ b/docs/reference/attribute.rst
@@ -11,6 +11,7 @@ Attribute
 .. autoattribute:: Attribute.dimension
 .. autoattribute:: Attribute.shape
 .. autoattribute:: Attribute.name
+.. autoattribute:: Attribute.extra
 
 .. toctree::
     :maxdepth: 2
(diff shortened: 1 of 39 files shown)


cprogrammer1994:
Thank you!

PR #241, opened 2018-08-11
Add more docs and examples to flags

einarf:
Clarify what ``enable_only`` does exactly and give some examples setting flags. It's easy to forget these are actual bit flags that can be ORed together.


cprogrammer1994:
Thank you!

PR #243, opened 2018-08-23
Bug: Only set compare func if valid enum

einarf:
### Tiny bug fix

Only set compare func when having a valid enum.
Trying to set depth func to 0 triggers INVALID_ENUM.


cprogrammer1994:
Thank you! :)

PR #244, opened 2018-08-25
[WIP] Use find_namespace_packages in setup

einarf:
Tweaking setup to use ``find_namespace_packages``.

May need tweaks for travis, appveyor and rtd..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ addons:
 
 install:
   - export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start
+  - python -m pip install -U setuptools
   - python -m pip install -r tests/requirements.txt
   - python setup.py build_ext --inplace
   - python -m pip install -e .
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,47 @@
+# Template from https://packaging.python.org/guides/supporting-windows-using-appveyor/
+environment:
+
+  matrix:
+
+    # For Python versions available on Appveyor, see
+    # http://www.appveyor.com/docs/installed-software#python
+    # The list here is complete (excluding Python 2.6, which
+    # isn't covered by this document) at the time of writing.
+    - PYTHON: "C:\\Python36-x64"
+
+install:
+  - "%PYTHON%\\python.exe -m pip install -U setuptools"
+  - "%PYTHON%\\python.exe -m pip install -r tests\\requirements.txt"
+  - "%PYTHON%\\python.exe setup.py build_ext --inplace"
+  - "%PYTHON%\\python.exe -m pip install -e ."
+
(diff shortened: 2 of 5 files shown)


einarf:
@cprogrammer1994 I have no idea how you configured appveyor. The only thing I'm trying to achieve is to upgrade setuptools before setup.py is called. I suspect I'm on the wrong path and you might have some custom config for the project.

einarf:
FYI: No hurry when it comes to resolving this one.

I'd consider this something fairly low priority, but wanted to at least share the setup here in a PR since I just started using ``find_namespace_packages` for my own projects.

cprogrammer1994:
@einarf I have configured appveyor on their site. We should use a config file instead. I will make this change as soon as I can. The problem is that `find_namespace_packages` is available in setuptools [v40.1.0](https://setuptools.readthedocs.io/en/latest/history.html)

Here is the [setuptools changelog](https://setuptools.readthedocs.io/en/latest/history.html)

I am not sure how can we add a dependency to this version of setuptools. Is there a solution for this?
What we could try to achieve is to get setuptools updated before our setup.py is called. (before install deps or similar?)

einarf:
It's just adding the upgrade to the ``install`` step (like in the example template in this PR)

```
install:
  - "%PYTHON%\\python.exe -m pip install -U setuptools"
  - "%PYTHON%\\python.exe -m pip install -r tests\\requirements.txt"
  - "%PYTHON%\\python.exe setup.py build_ext --inplace"
  - "%PYTHON%\\python.exe -m pip install -e ."
```

The exact same solution as in ``.travis.yml``

einarf:
If you need me to do more work in this one, just poke. Optionally just take what you need from the PR and close it. Coordinating this change might not be practical through a PR from me since you have better control over the configuration. 😄

cprogrammer1994:
Hi!

I think `find_namespace_packages` is convenient enough to be used.
I wanted to know if I can manage to get moderngl installed with a clean python install (where no upgrade was made to the setuptools yet)

Since we depend on almost the latest setuptools, somehow we will need it as a dependency.
To be honest I did not find the time to play around with it :(

cprogrammer1994:
I'm trying to do progress on the experimental branch. I think it is still not convenient enough for new developers to get familiar with it. I'm tweaking it so one day it will be easy to maintain+extend, while keeping the performance :)

So far I got it render some triangles, some tests can also be executed on travis :)

If you rebase this commit against the setup.py of the experimenta-2 branch I will definitely merge it.
I want this feature in the master as well, but first I want to know if it will break any install scripts having no setuptools upgrade before the install of ModernGL.

einarf:
oh. completely forgot about this one. I'm glad it was useful in some way at least 😄

PR #245, opened 2018-08-28
Depth texture incompleteness

einarf:
Set a sane default filtering for depth textures to avoid incompleteness issues. The default filtering would also then align with what is advertised in docs:
https://moderngl.readthedocs.io/en/stable/reference/texture.html#moderngl.Texture.filter

According to the OpenGL wiki the default min filter is ``GL_NEAREST_MIPMAP_LINEAR``:
https://www.khronos.org/opengl/wiki/Common_Mistakes#Creating_a_complete_texture

.. so this causes incompleteness issues unless explicitly set. Texture wrap defaults are all good.

### DIFF
--- a/src/Texture.cpp
+++ b/src/Texture.cpp
@@ -232,8 +232,8 @@ PyObject * MGLContext_depth_texture(MGLContext * self, PyObject * args) {
 
 	gl.BindTexture(texture_target, texture->texture_obj);
 
-	// gl.TexParameteri(texture_target, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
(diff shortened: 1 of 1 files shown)


einarf:
Since the lines was already in the code commented out, I'm not entirely sure if they were commented out for a good reason. Repeat doesn't really matter, but what about filtering?

Also, can you auto-squash this one on merge? 😄 
I messed up committing to my local master initially.

cprogrammer1994:
Looks ok to me. Thank you!
They were commented out because I did not know using depth textures with sampler2D uniforms is a core feature. (I thought it was part of the compatible profile)

Issue #62, opened 2016-07-17
gl2ps backend ?


stuaxo:
Any chance of supporting gl2ps ?

VTK manages to use this to output to pdf and svg, it would be great if ModernGL could support this for the same reason.

cprogrammer1994:
I will check out what is **gl2ps**

cprogrammer1994:
```
Could not find a version that satisfies the requirement gl2ps
```

On PyPI:

Index of Packages Matching 'gl2ps'

There were no matches.

cprogrammer1994:
```
gl2psBeginPage and gl2psEndPage                # must be ported
gl2psText and gl2psTextOpt                     # must be ported

gl2psDrawPixels                                # raster methods are deprecated
                                                 (must find an alternative)

gl2psSpecial                                   # not OpenGL dependent

gl2psEnable and gl2psDisable                   # implemented in ModernGL
gl2psPointSize and gl2psLineWidth              # implemented in ModernGL

gl2psBlendFunc                                 # implemented in ModernGL
                                                 (currently not customizable)

gl2psBeginViewport and gl2psEndViewport        # must be ported
gl2psSetOptions and gl2psGetOptions            # must be ported
```

cprogrammer1994:
One can render to framebuffers with ModernGL, the content can be read and stored in some common image file formats with 1 - 5 lines of additional code to the rendering part. Pillow is capable of saving EPS files as well.

stuaxo:
I think (though I might be wrong) that you get real vector output with gl2ps, whereas the frambuffer would be a bitmap.

cprogrammer1994:
In this case, the only advantage of using modern OpenGL is to apply a projection and/or transformation to the vertices than save them to a file. The compute shader or Transform shader is capable of transforming  the vertices. Do the GL2PS able to convert the output of the fragment shader to vertex attributes (for example lighting aplied on a sphere).
I think the per vertex attribute can be calculated using normals and color values and texturing can be done as well fron vertex shader. The capability to read (transformed) vertex buffers will solve the problem partly. The newly created vertex attributes shall be converted to a format that PS can understand.

stuaxo:
Yup, makes sense to close this for now,  - if I ever find out more I can try and revisit this with an actual patch :)

PR #248, opened 2018-09-02
Better feedback when shaders doesn't compile

yoyonel:
### Description

Improving  the default error message from moderngl with more feedback by adding some of the actual source in stdout.

From demosys-py issue: [Better feedback when shaders do not compile #62](https://github.com/Contraz/demosys-py/issues/62)

Example of new output for shader build errors:
```
GLSL Compiler failed

tess_control_shader
===================

  0001  
  0002                          #version 400 core
  0003  
  0004                          layout(vertices = 4) out;
  0005  
  0006                          void main() {
  0007                            // set tesselation levels, TODO compute dynamically
  0008                            gl_TessLevelOuter[0] = 1;
  0009                            gl_TessLevelOuter[1] = 32;
  0010                            
  0011                            l_TessLevelOuter[2] = toto;

error C1008: undefined variable "l_TessLevelOuter"
error C1008: undefined variable "toto"

  0012  
  0013                            // pass through vertex positions
  0014                            gl_out[gl_InvocationID].gl_Position = gl_in[gl_InvocationID].gl_Position;
  0015                          }
  0016                          

```

### List of changes

- **added** Program_Error.cpp tools for formatting shader information logs. New units tests in test_program_errors.py
- **removed** Nothing
- **fixed** Nothing
- **changed** Program.cpp to use this new feature

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         architecture: ['x64']
 
     steps:
-      - uses: actions/checkout@v2      
+      - uses: actions/checkout@v2
       - name: setup
         uses: actions/setup-python@v2
         with:
--- a/moderngl/src/Program.cpp
+++ b/moderngl/src/Program.cpp
@@ -1,6 +1,9 @@
 #include "Types.hpp"
 
+#include "Program_Error.hpp"
+
 #include "InlineMethods.hpp"
+
 
 PyObject * MGLContext_program(MGLContext * self, PyObject * args) {
 	PyObject * shaders[5];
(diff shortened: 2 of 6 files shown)


yoyonel:
Yep the result logs are pretty volatile depend of GPU family (and OpenGL Driver implementation).
I can't support all OpenGL drivers (i don't have all GPU and drivers on my computer(s) ;-)).
If you have some ideas to implement a more generic implementation of this feature, don't hesitate ;-)
Maybe a possibility to send/express a generic parser could be useful ... i will try to figure out how to implement a feature like this.
For a first attempt, we can accept this feature and using (GitHub) fix/pr to add manually other logs parser.

cprogrammer1994:
Can we rebase this against the experimental-2 branch?
I would like to have the following setup:

All the compiler / linker errors will be passed to a user configurable function that returns a string with the error message, or set the proper exception.

The default would be to have this "transformation" of the error log.
I would like to let the user choose if he wants the "raw" error log.

I thought it could be configured in the following way.

```
ctx = moderngl.context(...)
ctx.set_glsl_error_formatter(...)  # to get the raw output set it to None
```

I think the the implementation could be in python. We should decide what parameters it could take. and what should it return / raise.

What kind of OS are you using? Can you build the experimental-2 branch?

Please let me know if not.

yoyonel:
I'm in vacation (portugal) but i will check tonight if i can build the experimental-2 branch (and merge it). (I'm on notebook ubuntu-mate (16.04) with gcc ... dont remember the version ^^)

cprogrammer1994:
the experimental-2 should build on linux at the moment. I use travis to see build errors on linux.

yoyonel:
Do you have some documentation about this branch (experimental-2) ?
The 'tests'/example folder are very ... slim :p

einarf:
I'll see if I can revive this at some point, but not for 5.6.

yoyonel:
@einarf i'v rebased this MR with `master` (from `moderngl`).
I had to deactivate (skip) utest on vertex shader error, because on CI (VM used by Github Action), this utest return strange error message (the error message is raw, not interpolated). the others utests work ... it's not perfect but it is what it is.
Moving the formatting message from the C part to Python could be a good thing, but it's more difficult (and more distant from the current `master` of moderngl). Maybe, i will try in other (next) step.

szabolcsdombi:
We will definitely have an option to add custom error parsers to moderngl.
Thank you very much for this pull request.

it will work similarly to https://github.com/szabolcsdombi/zengl/blob/main/examples/custom_shader_errors.py

szabolcsdombi:
I am closing this as the implementation is in C++,
Users will be able to hook the error generation and extend it or pretty format it like in the code you provided.
there will be an example or an extra modergl module providing something basic.
By default the original GL provided message will be shown.

PR #252, opened 2018-10-15
add compute shader example with median_5x5 shader

minuJeong:
### Description
Hello Szabolcs,

I found modernGL was useful for my study, I'd like to make a contribution.

I spent some time to searching reference of using compute shader, and maybe this could help others to get start with it faster with this example.

(Wait for loading gif file here)
image

### List of changes

- **added** ...
examples/08_compute_shader.py
examples/gl/median_5x5.gl

- **removed** ...
- **fixed** ...
- **changed** ...
nothing

### DIFF
--- a/examples/08_compute_shader.py
+++ b/examples/08_compute_shader.py
@@ -1,0 +1,79 @@
+
+'''
+    example of using compute shader.
+
+    requirements:
+     - numpy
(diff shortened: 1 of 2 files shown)


cprogrammer1994:
Thank you!

PR #254, opened 2018-10-29
add example 09_ssbo

minuJeong:
### Description
Add another example for compute shader.
If you run this code, it will popup a simple PyQt5 window with QOpenGLWidget, and renders hundreds of balls bouncing with wall. Advancing and bouncing ball calculations are done with compute shader using ssbo.

### List of changes

- **added** ...
example 09_compute_shader_ssbo


cprogrammer1994:
Thank you!

PR #261, opened 2018-12-11
Update pygame example

Vermeille:
Hi!

I updated the pygame example to better match ModernGL 5.
I somehow had to move to GL ES 3.10 to please my computer so change that if needed.

Have a nice day


cprogrammer1994:
Thank you! :)

PR #264, opened 2019-01-07
State requested and acquired gl version

einarf:
When creating a context from an existing opengl context we should state the requested and aquired opengl version to make the error message more clear.

### Description

Stumbled over this tiny detail today. We do this for standalone context, but not when using existing contexts. It make it much easier to understand when context creation fails for various reasons.


cprogrammer1994:
Thank you!

PR #271, opened 2019-01-16
Example windows (Completed)

einarf:
Examples using ModernGL with pyqt5, pyglet, glfw and sdl2

Each window type support the folloing features:

* Runtime resizing and viewport update
* Fullscreen mode
* Multisample antialiasing (MSAA) default at 4
* Keyboard events
* Mouse position and button events
* Vsync toggle
* Toggle mouse cursor on or off

Events are passed from the window to an example class in a platform and framework-indenpedent way.

## Example Options

```
optional arguments:
  -h, --help            show this help message and exit
  -w {glfw,pyglet,pyqt5,sdl2}, --window {glfw,pyglet,pyqt5,sdl2}
                        Name for the window type to use
  -fs, --fullscreen     Open the window in fullscreen mode
  -vs VSYNC, --vsync VSYNC
                        Enable or disable vsync
  -s SAMPLES, --samples SAMPLES
                        Specify the desired number of samples to use for
                        multisampling
  -c CURSOR, --cursor CURSOR
                        Enable or disable displaying the mouse cursor
```

For example:

```bash
# Run example in fullscreen with 8 x MSAA, vsync and no visible mouse cursor
python 03_alpha_blending.py --fullscreen --samples 8 --cursor off --vsync on

# Run example using SDL2
python 03_alpha_blending.py --window sdl2

# Run example using pyglet
python 03_alpha_blending.py --window pyglet

# Run example using glfw
python 03_alpha_blending.py --window glfw
```

## Example Template

```py
from window import Example, run_example


class MyExample(Example):
    # The minimum required OpenGL version (default 3.3)
    gl_version = (3, 3)
    # Title of the window
    title = "Empty Window"
    # Size of the window to open
    window_size = (1280, 720)
    # Lock aspect ratio letting the window class automatically
    # calculate the right viewport regardless of window size.
    # This can add black borders because the viewport is centered
    # vertically and horizontally.
    aspect_ratio = 16 / 9
    # Should users be able to resize the window?
    resizable = True

    def __init__(self, **kwargs):
        # ctx and wnd are mainly passed in there
        super().__init__(**kwargs)

        # The ModernGL context is avaialble
        # self.ctx.something()

        # A window reference is also availabel
        # self.wnd.size

    def render(self, time: float, frame_time: float):
        """
        Called every frame

        Params:
            time (float): Number of seconds elapsed since rendering started
            frame_time (float): Seconds since last frame
        """
        self.ctx.clear(0.2, 0.4, 0.7)
        # Render stuff here

[...]

PR #273, opened 2019-02-03
Replaced obsolete QGLWidget with QOpenGLWidget

LexKex:
### Description
Replaced an obsolete QGLWidget class.

From: http://doc.qt.io/qt-5/qglwidget.html

"This class [QGLWidget] is obsolete. It is provided to keep old source code working. We strongly advise against using it in new code."

### List of changes

- replaced QGLWidget with QOpenGLWidget class
- replaced QGLFormat infrastructure with QSurfaceFormat infrastructure


einarf:
FYI: I'm in the middle of rewriting all example windows in #271

PR #276, opened 2019-02-10
Example: Simple Camera Class

LexKex:
### Description

A stand-alone camera example. 

### List of changes

- Added example "11_simple_camera_controls.py"
- The example uses the canon "example_window.py" for context creation
- The camera class is built using following tutorials:
          https://learnopengl.com/Getting-started/Camera
          http://in2gpu.com/2016/03/14/opengl-fps-camera-quaternion/


cprogrammer1994:
Hi! Thank you for the new example. I want to merge it, but it seems you have based your example on the previous PR of yours. (I mean we also have the example_window changes)

Does your example depend on the changes in 7c8e319e68074cce7b0f4?
If not, please revert 7c8e319e68074cce7b0f4 so #271 will not conflict.

PR #278, opened 2019-02-11
A new example: camera controls

LexKex:
### Description

A stand-alone camera example.

### List of changes

* Added example "11_simple_camera.py"
* The example uses the canon "example_window.py" for context creation
* The camera class is built using following tutorials:
     https://learnopengl.com/Getting-started/Camera
     http://in2gpu.com/2016/03/14/opengl-fps-camera-quaternion/


cprogrammer1994:
Thank you!

einarf:
I noticed the camera controls don't work on OS X (and probably not Linux) because it checks OS specific key IDs (a trap I have fell into in the past as well). I'll fix that when converting it to the new example system.

PR #281, opened 2019-02-21
Convert 11_simple_camera to new example system

einarf:
That's about it. I had to simplify the controls a bit because the key inputs are a bit more limited considering they are supporting multiple libraries. Something to look into in the future.


cprogrammer1994:
Thank you!

PR #285, opened 2019-02-23
Example window polish

einarf:
I went through code and did more extensive testing:

* **pyqt5**: Entering fullscreen mode was slow and glitchy. Should be a lot more snappy now.
    * Don't try to center a fulllscreen window. This caused the window to glitch for a moment
    * Only call `windget.show()` in fullscreen mode. We only need to call `showFullscreen()`. This avoids sudden flickering in the first couple of seconds after the window is created.
* **glfw**: Match the exact video mode of the desktop to avoid video mode switching when entering fullscreen
* **pyglet**: Use the current desktop size in fullscreen mode to avoid video mode switching
* **Bug**: Not all windows called `resize` on the example instance
* **Bug**: Always report `buffer_size` to the example instead of the window size
* **Improvement**: Remove redundant `close` and `is_closing` members and rely on base class when possible
* **Improvement**: Mention supported platforms in `README`
* **Improvement**: Add more comments and fix spotted typos

I think ModernGL example windows are actually better than the ones in demosys-py. I need to do somthing about that soon 😄

### DIFF
--- a/examples/00_empty_window.py
+++ b/examples/00_empty_window.py
@@ -14,6 +14,13 @@ class EmptyWindow(Example):
     def render(self, time: float, frame_time: float):
         self.ctx.clear(0.2, 0.4, 0.7)
 
+    def resize(self, width: int, heigh: int):
+        """
+        Pick window resizes in case we need yo update
+        internal states when this happens.
+        """
+        print("Window resized to", width, heigh)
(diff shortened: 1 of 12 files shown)


cprogrammer1994:
Thank you :)

PR #287, opened 2019-02-23
docs: 'pip install' commands should use lower case 'moderngl'

tartley:
### Description

The docs and examples contain a few 'pip install' commands that show the older:

    pip install ModernGL

I changed them to use the latest lower-cased:

    pip install moderngl

I'm just assuming that all the docs and examples still work correctly on v5.x. It could be that I've updated things that should have remained as they are. 

Thank you for the invitation to add myself to the README but my contributions are all so tiny that I don't feel worthy. Hugs!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 # ModernGL
 
 ```sh
-pip install ModernGL
+pip install moderngl
 ```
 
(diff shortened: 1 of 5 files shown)


cprogrammer1994:
Than you! It is more consistent to install and import "moderngl"

PR #293, opened 2019-02-25
Complete sampler implementation

cprogrammer1994:
### List of changes

- **added** "missing things" of the sampler objects


cprogrammer1994:
```py
ctx.sampler(self, texture, filter=0x2600, wrap=0, anisotropy=1.0, compare_func=None, lod_range=(-1000, 1000), lod_bias=0.0, border=0.0)

sampler.filter = ...  # type: Tuple[int, int]
sampler.wrap = ...  # type: int
sampler.anisotropy = ...  # type: float
sampler.compare_func = ...  # type: str
sampler.lod_range = ...  # type: Tuple[int, int]
sampler.lod_bias = ...  # type: float
sampler.border = ...  # type: Tuple[float, float, float, float]
sampler.texture = ...  # type: Texture

sampler.use(location=0)
```

Issue #306, opened 2019-03-08
document format string used by 'content' arg of vertex array constructor


tartley:
I plan to contribute this documentation. I hijacked another existing issue discussing it. I'll discuss it here instead.

tartley:
I would like to contribute some documentation.

I am trying to figure out the format strings passed to `Context.vertex_array`, and when I've figured it out, I'd be happy to add a doc PR explaining my findings. :-)

For example, I'm talking about the '2f' in:

```
ctx.vertex_array(
            self.shader,
            [
                (self.ctx.buffer(verts), '2f', 'vert'),
                (self.ctx.buffer(colors), '3u', 'color_in'),
            ],
            self.ctx.buffer(indices),
            index_element_size=1,
        )
```

I don't know whether these format strings are using a syntax from elsewhere, eg that maybe is part of OpenGL? If so, then the docs could link to those docs. Or if it is just made up for moderngl, then the docs could describe the format.

At first I assumed that it was the same type of format strings that I pass to `struct.pack(...)` to create my buffers, because '3f' does the same thing in each. But this doesn't seem to be the case. In particular, I'm trying to create a buffer of colors made up of three unsigned bytes. I create these using `struct.pack('3B', *data)`, but vertex_array constructor does not accept '3B' as a valid format string.

Just ten minutes ago, I found the source that reads these format strings, here:
https://github.com/cprogrammer1994/ModernGL/blob/master/src/BufferFormat.cpp

So now I can see what sequences of values are allowed. At the very least, I could try to lay out this information in a readable manner for the documentation. Would this be useful? If you have strong opinions about how it should look, let me know.

---

In my specific case, to pass in vertex color attributes using unsigned bytes, and receive them as floats from 0.0 to 1.0 in my vertex shader.  I think it's important for me to get `node.normalize` set to true. From the source of Bufferformat.cpp, it appears the only way to do this is to use a format like '3f1'. So I'm puzzled about why 'f' is used here for unsigned bytes. I've been assuming 'f' is shorthand for 'float'. Any clues welcome. **Update:** yay using f1 works.

tartley:
One option would be to use BNF, eg. a hurried version:

```
<nodes> ::= <node> | <node> " " <nodes>
<node> ::= <opt-count><type-size><opt-post>
<opt-count> ::= <integer> | ""
<type-size> ::= ( "f" <size8> ) | ( "i" <size4> ) | ( "u" <size4> ) | ( "x" <size8> )
[...]

PR #309, opened 2019-03-13
Documentation for buffer formats.

tartley:
### Description

Issue: https://github.com/moderngl/moderngl/issues/306 document format string used by 'content' arg of vertex array constructor
 
Changes:

1. Add new "_Buffer Format_" documentation page, as docs/reference/buffer_format.rst
2. Add a link to it from docs/reference/index.rst
3. Add a link to it from docstring of Context.vertex_list
4. Also do (1) and (2) in the 'next' documentation.
5. The 'next' version also notes that "3f1" can be passed to a "vec4", as well as a "vec3".

### Status

Review requested. I've done local `python setup.py build_sphinx` and it looks good to me.

All review comments thus far are implemented.

### Questions for reviewers

(ie TODOs that I don't currently intend to do anything about)

I have linked to this new page from `Context.vertex_array`. Is there anywhere else I should also link from?

I just discovered `test_documentation.py`. My reading of it is that it's only for documenting the API, not for stand-alone handwritten doc pages like this. So I have nothing to add here. Is that right?

I propose postponing the documentation of dtypes to a separate issue. I've collected all the hints and overviews you wrote on a new issue, here: https://github.com/moderngl/moderngl/issues/314

The thing I'm least happy about with this PR is that it's turned out to be more wordy than I'd hoped for. If your honest opinion is something like "_couldn't we write the same info in about half as many words?_", let me know and I'll happily rethink it.

I've described the trailing `\v \i \r` suffixes as a "_usage_". If there's a better term, please let me know.

In general, I'm still getting up to speed on modern OpenGL terminology, so may have used the wrong words in many places.

---

Forgive me, I'm not adding the changed files here as requested by the PR template, git does that for us. Please shout at me if this is unhelpful.


tartley:
I haven't done anything about linking to this page from elsewhere in the docs. I can see that Context.vertex_array should link here. Anywhere else?

einarf:
Read through it a couple of times. Overall it looks very detailed and accurate. Well done! I might add some comments on specific details.

docs/reference/buffer_format.rst
```diff
@@ -0,0 +1,264 @@
...
+or once per instance. Using ``/r``, the buffer only needs to contain one
+color value, which is passed to every vertex of every instance for the whole
+render call.
+
+An alternative approach would be to pass in the color as a uniform, since
+it is constant. But doing it as an attribute is more flexible. It allows us to
+reuse the same shader program, bound to a different buffer, to pass in color
+data which varies per instance, or per vertex.
+
+Another alternative approach would be to not specify the color in our render at
```
einarf: It this paragraph necessary?

einarf:
> I have linked to this new page from Context.vertex_array. Is there anywhere else I should also link from?

I think the current links are sufficient.

> I just discovered test_documentation.py. My reading of it is that it's only for documenting the API, not for stand-alone handwritten doc pages like this. So I have nothing to add here. Is that right?

Correct. It's mainly there to verify that all attributes/methods are actually documented correctly.

> I propose postponing the documentation of dtypes to a separate issue. I've collected all the hints and overviews you wrote on a new issue, here: #314

Perfectly reasonable.

> The thing I'm least happy about with this PR is that it's turned out to be more wordy than I'd hoped for. If your honest opinion is something like "couldn't we write the same info in about half as many words?", let me know and I'll happily rethink it.

I see what you mean, but at the same time it make the docs precise. Maybe in the example section you can instead show a snippet/header of the corresponding glsl shader. This means that the quicker readers can see the correlations between the code blocks quickly.

> In general, I'm still getting up to speed on modern OpenGL terminology, so may have used the wrong words in many places.

I think the wording in the PR makes sense.

> Forgive me, I'm not adding the changed files here as requested by the PR template, git does that for us. Please shout at me if this is unhelpful.

Agreed 😄

tartley:
@einarf Thanks for the review. I'll tidy up the things you hinted could be changed.

Also: Should make the 'next/reference' changes too? Or is there a planned separate process for updating 'next' docs?

einarf:
> Also: Should I make the 'next/reference' changes too? Or is there a planned separate process for updating 'next' docs?

Definitely add apply the changes in `next` as well. I think the plan is to publish docs from both locations and when 6.x is the default they are moved down.

tartley:
I'm all done. Changes I made:

* Add snippets of shader attribute declarations to the examples.
* Apply the same changes to the 'next' docs as well, but there are no docstrings in "next/context.py", so I didn't link to it from there.

As far as I know, this is ready for merge.

docs/reference/buffer_format.rst
```diff
@@ -105,23 +105,29 @@ Example buffer formats
 ......................
 
 ``"2f"`` has a count of ``2`` and a type of ``f`` (float). Hence it describes
-two floats. The size of the floats is unspecified, so it defaults to ``4``. The
-usage of the buffer is unspecified, so defaults to ``/v`` (vertex), meaning
-each successive pair of floats in the array are passed to successive vertices
-during the render call.
+two floats, passed to a ``vec2`` shader variable. The size of the
```
einarf: "attribute" is more accurate. "variable" is not consistent with GLSL terms. I think there are couple of them in the doc. We also refer to shaders as "programs" in monderngl, but I guess since attributes are present in the vertex shader it's probably still accurate to use "shader".

einarf:
Found one tiny thing. Other than that, this looks great!

einarf:
Also, don't forget to add yourself to contributor list if you want to be there 😄

docs/next/reference/buffer_format.rst
```diff
@@ -0,0 +1,283 @@
...
+=============
+
+.. py:module:: moderngl
+.. py:currentmodule:: moderngl
+
+Description
+-----------
+
+A buffer format is a short string describing the layout of data in a vertex
+buffer array (VBO).
```
cprogrammer1994: vertex buffer array -> vertex buffer object (vbo)

docs/next/reference/buffer_format.rst
```diff
@@ -0,0 +1,283 @@
...
+   - ``u`` unsigned int
+   - ``x`` padding
+- ``size`` is an optional number of bytes used to store the type.
+  If omitted, it defaults to 4 for numeric types, or to 1 for padding bytes.
+
+  A format may contain multiple, space-separated ``[count]type[size]`` triples
+  (See the :ref:`example-of-single-interleaved-array-label`), followed by:
+
+
+- ``/usage`` is optional. It may be preceded by a space, and then consists of
```
cprogrammer1994: may -> should

In moderngl 6 it will be a must. It caused a confusion because on the first sight one may think it belongs only to the last element.

docs/next/reference/buffer_format.rst
```diff
@@ -0,0 +1,283 @@
...
+   - ``/v`` per vertex.
+     Successive values from the buffer are passed to each vertex.
+     This is the default behavior if usage is omitted.
+   - ``/i`` per instance.
+     Successive values from the buffer are passed to each instance.
+   - ``/r`` per render.
+     the first buffer value is passed to every vertex of every instance.
+     ie. behaves like a uniform.
+
+  When passing multiple VBOs to a VAO, the first one must be of usage ``\v``,
```
cprogrammer1994: `\v` -> `/v`

cprogrammer1994:
I think this documentation is great. I will merge this soon. I have added some notes, but in case you don't want to change them, please respond to them. I will merge this next week and add small changes if necessary. I really want this PR to be part of the next release, it will be live on rtd.
Thank you for your contribution.

tartley:
Thank you both for reviewing this. I've implemented all review comments.

One issue that I'm still dithering about, I'll describe here in case you have strong opinions:

The issue of "3f1" being assignable to a vec4 was initially only described in the v6 docs. But then, adding subsequent commits, it was a hassle to keep the v5 and v6 doc in sync. I should probably have figure the command to apply patches to both files, but instead I was manually editing each.

So, I decided to put that para into BOTH files, but qualified with "in ModernGL v6.0". This means that the v5 page and v6 page were identical, which made it easy to keep them in sync.

Later, I noticed building the docs gave errors because I had duplicate RST labels (eg. used for linking from the description to the relevant examples.), such as:

```
.. _example-of-simple-usage-label:
```

which exists in both documents. So I changed the labels in the v6 directory to look like:

```
.. _example-of-simple-usage-label-v6:
```

This silences the doc build warnings, but obviously means the two docs are no longer identical. That's how I've left it.

docs/next/reference/buffer_format.rst
tartley: fixed

docs/next/reference/buffer_format.rst
tartley: fixed

docs/next/reference/buffer_format.rst
tartley: fixed

PR #312, opened 2019-03-14
Apply index_element size to pointer offsets in DrawElements

einarf:
Doing PR so we don't accidentally conflict.


cprogrammer1994:
Than you!

I want to release 5.x soon because of this fix.

Issue #367, opened 2020-01-10
Share between context


DarwinSenior:
I was wondering if it is possible to share textures between contexts

As afar I know there is no way to create a texture from an existing texture right?

**Describe the solution you'd like**

Is it possible to do something like this?

```python
# setup the first window ...
ctx1 = moderngl.create_context()
texture1 = ctx1.texture((H, W), 4)

# setup the second window ...
ctx2 = moderngl.create_context()
texture2 = ctx2.texture(texture1)
frame = ctx2.framebuffer((texture2,))
```
Where `ctx2.texture` function will check if `texture1` can be shared and create a texture that is shared between `ctx1` and `ctx2`.

**Version info**

```
moderngl 5.5.4
--------------
vendor: NVIDIA Corporation
renderer: GeForce GTX 1080 Ti/PCIe/SSE2
version: 4.6.0 NVIDIA 430.64
python: 3.6.9 (default, Nov  7 2019, 10:44:02)
[GCC 8.3.0]
platform: linux
code: 460
```

einarf:
Another option is to make different windows sharing the same context. That might be simpler, but I don't know about your use case. 

Some notes on this for glfw:
https://www.glfw.org/docs/latest/context_guide.html#context_sharing

einarf:
Does your example assume object sharing between contexts? If you have any more details, that would be great.

DarwinSenior:
I do ask if it is possible different windows sharing objects between contexts. 
So, what I had in mind is that no object is created in the new context as it is managed by the original context but can be used in the new object.

What I want to achieve is that we can create a "borrowed" object (buffer/texture/storage) by an id (GLuint), so that they can be used with the rest of the moderngl code. Basically to create same object but with a different context. As I understand it, textures, buffer objects, and render buffers are the ones can be shared between contexts.

For sharing between context, one scenario I currently have is to that I want to render the same content to two screens, where we want to output to one screen a rendered 3D scene and the other screen a tone-mapped version of the same content. In this case, it makes perfect sense to render to a buffer first and then blit this texture to one screen and process the same texture in another.

einarf:
This is close to what I imagined as well. I assume if enabling object sharing we could test reconstructing the moderngl object in the other context. I have to look into how this works. If we object sharing means name sharing the moderngl texture instance could be reconstructed from a `GLunit`.

Currently I don't think this is possible in moderngl. We would have to come up with some solution for the next release (5.6).

There are still a couple of ways you can do this:

[...]

PR #370, opened 2020-01-20
Adding Texture transfer

DarwinSenior:
### Description

Trying to pull with issue #367.
All data types (buffers/textures) can be sharable between contexts.

Adding `texture.transfer(ctx)` to share between context

### List of changes

- **added** ...
`Texture.transfer`
`Texture3D.transfer`
`TextureCube.transfer`
`TextureArray.transfer`
`Buffer.transfer`

I have listed my changes. However, it is only a temporary solution.
A problem I could see with this solution comes from synchronisation.

A better solution is to create a new type that refers to the original object and operates
on that object.

### DIFF
--- a/moderngl/buffer.py
+++ b/moderngl/buffer.py
@@ -270,3 +270,20 @@ class Buffer:
             (self, index) tuple
         """
         return (self, index)
+
+    def transfer(self, context):
+        '''
+            share the buffer to a new context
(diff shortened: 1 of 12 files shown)


einarf:
We found out this is not necessary. Be free to reopen if this is not the case.

PR #381, opened 2020-02-01
easily understand

zhouhang95:
The old code is hard to understand.


einarf:
Thanks! It seems reasonable.

PR #387, opened 2020-02-19
Fix catching a context creation error

asnt:
With the switch tot glcontext, it seems a base `Exception` is raised from `moderngl.create_context()` instead of a `moderngl.Error` previously.


einarf:
Thank you!

Issue #390, opened 2020-03-20
Crash loading compute shader when too low opengl version


einarf:
Not a super critical issue, but we should at least do a sanity check during compute shader creation and give some reasonable feedback.

einarf:
Like mentioned in the PR we should check version code AND extensions. I don't really want to add this before more research is done. We might break things.

einarf:
This is someting we could do in the refactoring on the python side.

```py
supported = ctx.version_code >= 430 or "GL_ARB_compute_shader" in ctx.extensions
```

My only worry is that some drivers might only have the vendor specific extension.

szabolcsdombi:
the list should be finite and I expect having a custom entry but no `GL_ARB_compute_shader` will be rare.
these will be reported, the bugfix is trivial and requires only python code changes. it even can be tested without rebuilding moderngl.

szabolcsdombi:
This was not implemented as loading and searching the extension list is an expensive operation. load time would suffer from it.
I think we should just tie `glDispatchCompute != NULL` to having compute shaders available.

PR #396, opened 2020-05-01
fix for issue #390

Leterax:
Check OpenGL version before creating a Compute-Shader.

### Description

Version is now checked before creating a Compute-Shader

### List of changes

- **added** Version check
- **removed** ...
- **fixed** ...
- **changed** ...

### DIFF
--- a/moderngl/context.py
+++ b/moderngl/context.py
@@ -1268,7 +1268,10 @@ class Context:
             Returns:
                 :py:class:`ComputeShader` object
         '''
-
+        
+        if self.version_code < 430:
+          raise ValueError("OpenGL version 430+ needed, found {}".format(self.version_code))
+  
         res = ComputeShader.__new__(ComputeShader)
         res.mglo, ls1, ls2, ls3, ls4, res._glo = self.mglo.compute_shader(source)
 


Leterax:
The error message is probably too vague, but I want sure how to word it correctly.

einarf:
Not 100% if this is the right way to do it. We might need to actually check if the compute shader extension is available. Need to look into this,

Leterax:
It seems you can get away with using ARB_compute_shader <https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_compute_shader.txt>, but this requires OpenGL 4.2. 
So we can be sure they arent supported below 4.2 but need to check for ARB_compute_shader if the version is below 4.3.

Leterax:
after some googling I found this:
http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-12-opengl-extensions/
they suggest this to find if an extension is supported:
```c
int NumberOfExtensions;
glGetIntegerv(GL_NUM_EXTENSIONS, &NumberOfExtensions);
for(i=0; i<NumberOfExtensions; i++) {
  const GLubyte *ccc=glGetStringi(GL_EXTENSIONS, i);
  if ( strcmp(ccc, (const GLubyte *)"GL_ARB_compute_shader") == 0 ){
    // The extension is supported by our hardware and driver
  }
}
```

einarf:
Yup. Just need to make sure the context loads a `set` of extension strings that could be checked.

Leterax:
would that then be saved in the actual MGLContext struct?

einarf:
Yup. Just need to add it so extensions can be queried on the C++ and python side, It's available in C++ only. Not very practical. Doing a `extension in extensions` check on a `set` in python should be fast enough.

Leterax:
yeah. set contains checks are plenty fast enough. I have no idea what I'm doing, but I would imagine something like this would work:
```cpp
PyObject* extensions = PySet_New(NULL);
int NumberOfExtensions;
glGetIntegerv(GL_NUM_EXTENSIONS, &NumberOfExtensions);
for(i=0; i<NumberOfExtensions; i++) {
  const GLubyte *ccc=glGetStringi(GL_EXTENSIONS, i);
  PySet_Add(extensions, PyUnicode_FromString(ccc));
}
```
I'm not entirely sure if you can just directly convert GLubyte* to a unicode string, because a null byte terminated char buffer is expected.

Leterax:
PySet_New can also take an iterable, so it might be possible without the for loop adding to the set.

einarf:
`Context` just needs a pre-loaded set of extenstions that are created during context creation or something.

einarf:
Need to do some tests before deciding on this one. It might actually be correct using version in the context because this indicates the application wanted to add this limitation.

einarf:
After looking more into this: Compute shaders are supported if either the version code is high enough OR there is a known compute shader extension available. This PR only covers the first part. I'm not entirely sure what all the valid compute shader extension strings are.

What we can do it add some flags in the `Context` itself.

einarf:
We'll add a check after the refactoring. It shouldn't be too hard. Closing this PR and commenting in the original issue.

PR #429, opened 2020-12-10
Update _example.py

SwaySZ:
### Description

Examples work fine on MacOS, but fail on Ubuntu 18.04 with the following issue.
ubuntu pyglet window.nosuchconfigexception

### List of changes

remove "samples = 4" to fix the issue


SwaySZ:
fixed

einarf:
It's a good point that we should not have msaa enabled by default. I'll just remove this line for now. We don't need to merge in this comment. Thanks for pointing this out!

Issue #428, opened 2020-12-09
Simple example loading an image?


DanielArnett:
[I write fragment shaders for modifying images](https://github.com/DanielArnett/360-VJ/releases), and I'd like to port these to moderngl. I can't find an example that does what I need. 

I just need to get an image and input uv into a fragment shader, and return the resulting image. Does anyone know of an example like this or could help me with one?

einarf:
This was something that was recently discussed in our discord server. Probably a better place for questions like this. Some quick and dirty mockup code : https://paste.pythondiscord.com/janicujuro.py

Discord : https://discord.gg/UEMtW8D

DanielArnett:
Okay I made a few tweaks and submitted it to the examples folder in [this PR](https://github.com/moderngl/moderngl/compare/master...DanielArnett:patch-1)

DanielArnett:
Thanks for your help and the discord invite!

PR #430, opened 2020-12-10
Add an example for modifying an image using a shader.

DanielArnett:
[Resolves Issue 428](https://github.com/moderngl/moderngl/issues/428)

### Description

Write here.

### List of changes

- **added** An example that loads an image, applies a simple fragment shader, and saves the image.


einarf:
Thanks. Might modify this a bit more, but this is a good start.

einarf:
It's a good point that we should not have msaa enabled by default. I'll just remove this line for now.

PR #433, opened 2020-12-20
Fix color support and add colors to cairo integration example

numberZero:
### Description

Color support was broken because Cairo and OpenGL use different channel orders, RGBA and BGRA correspondingly (bytewise, on a little-endian system; despite the fact colors are commonly set as RGBA when specified componentwise in Cairo). The Cairo example was rendering grayscale only so the issue was not immediately visible, but when I tried to add colors it became obvious.

### List of changes

- **fixed** color support in the cairo integration example
- **changed** circle color in that example


einarf:
Thanks! I only tested with some white geometry.

einarf:
Swizzle is the way to go.

Issue #438, opened 2021-01-23
ModernGL is using glProgramUniform (4.1 feature) causing issues on 3.3 only


abey79:
**Description**

ModernGL crashes on Windows 10 running in Parallel Desktop 16 virtual machine. The root cause is the use of `glProgramUniformX()`, which is not part of OpenGL 3.3 (spec: OpenGL 4.1 and later) and is thus not implemented by Parallels Desktop (they support OpenGL 3.3).

**Proof of code**

The following line leads to a NULL pointer execution:

```python
ctx = moderngl.create_standalone_context()
prog = ctx.program("...", "...")
prog["var"].value = 1
```

Here is an illustration of the root cause:

```python
import glcontext

backend = glcontext.default_backend()
ctx = backend(mode='standalone', glversion=330)  #, libgl='libGL.so.1')
print(ctx)
print(ctx.load('glProgramUniform1iv'))
```
Result:
```
<wgl.GLContext object at 0x000001F54F8411E0>
0
```

**Version info**

```
moderngl 5.6.2
--------------
vendor: Parallels and ATI Technologies Inc.
renderer: Parallels using AMD Radeon Pro 560X OpenGL Engine (Compat)
version: 3.3 ATI-3.10.18
python: 3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]
platform: win32
code: 330
```

Parallels Desktop  16.1.2

einarf:
We should at the very least add something about this in docs. The `glProgramUniform*` functions are nice because you specify what program to set the uniform on while the `glUniform*` functions sets uniforms on the bound program. This makes the absraction moderngl currently has a lot more complicated.

Most drivers even just supporting 3.1/4.0 do also support `glProgramUniform`. It's fairly rare this is not the case.

abey79:
If it doesn't exist already, there should be a check function that should test for GL 3.3 + all required, non-GL 3.3 calls, so client code can gracefully exit when ModernGL will not function.

einarf:
Yup. With the new `ctx.extensions` we can at least do something here. I'm looking into it.
During context initialization we can check if `GL_ARB_separate_shader_objects` is supported.

Could also be possible to make an alternate path when setting up uniform getters and setters if the extension is not supported. It will probably have a performance impact when not supported, but at least it will work. It means we need to bind the program before setting the uniform and use `glUniform*` instead.

abey79:
[...]

PR #443, opened 2021-02-14
[Draft] Use glUniformX instead of glProgramUniformX

abey79:
### Description

As per #438, the use of `glProgramUniformX` is unsupported by some strict 3.3 implementations. This PR is my (very naive) attempt at addressing the issue by replacing all calls to `glProgramUniform` by the following sequence:

```c
GLint current_program;
gl.GetIntegerv(GL_CURRENT_PROGRAM, &current_programm);
gl.UseProgram(self->program_obj);
gl.UniformX(...);
gl.UseProgram(current_program);
```

This version compiles but trashes the display.

**Edit**: the PR now appears to work with my cursory testing.

### DIFF
--- a/moderngl/src/Types.hpp
+++ b/moderngl/src/Types.hpp
@@ -351,7 +351,6 @@ struct MGLUniform {
 
 	MGLProc value_getter;
 	MGLProc value_setter;
-	MGLProc gl_get_integer_proc;
 	MGLProc gl_use_program_prog;
 	MGLProc gl_value_reader_proc;
 	MGLProc gl_value_writer_proc;
--- a/moderngl/src/Types.hpp
+++ b/moderngl/src/Types.hpp
@@ -351,6 +351,8 @@ struct MGLUniform {
 
 	MGLProc value_getter;
 	MGLProc value_setter;
+	MGLProc gl_get_integer_proc;
+	MGLProc gl_use_program_prog;
 	MGLProc gl_value_reader_proc;
 	MGLProc gl_value_writer_proc;
 
--- a/moderngl/src/Uniform.cpp
+++ b/moderngl/src/Uniform.cpp
@@ -45,11 +45,17 @@ int MGLUniform_set_data(MGLUniform * self, PyObject * value, void * closure) {
 		return -1;
 	}
 
+	GLint current_program;
+    ((gl_get_intergerv_proc) self->gl_get_integer_proc)(GL_CURRENT_PROGRAM, &current_program);
+    ((gl_use_program_proc)self->gl_use_program_prog)(self->program_obj);
+
 	if (self->matrix) {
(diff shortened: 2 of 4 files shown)


cprogrammer1994:
Hi,

Back when we added `glProgramUniform*` over `glUserProgram`+`glUniform*` we were not able to find drivers with OpenGL3.3 support but no `glProgramUniform*` from extensions.

I suggest to make `glProgramUniform*` functions in C++ and load those if the original ones are not available by the driver.
The `glProgramUniform` implementation is just a glUseProgram + glUniform* (also check if the bound program must be restored)

I think this way you should not bother checking if the changes break existing code due to the NULL functions were used before patched.
Also we do not affect users who had the `glProgramUniform*` available

What do you think?

einarf:
> I suggest to make `glProgramUniform*` functions in C++ and load those if the original ones are not available by the driver.
> The `glProgramUniform` implementation is just a glUseProgram + glUniform* (also check if the bound program must be restored)
> 
> I think this way you should not bother checking if the changes break existing code due to the NULL functions were used before patched.
> Also we do not affect users who had the `glProgramUniform*` available

Yep. That's what I was thinking as well. This is just a quick experiment to get moderngl running on Parallels for MacOS. With the exception of raspberry pi it's the only place I've seen a 3.3 version not supporting the separate program extension.

abey79:
@cprogrammer1994 I totally agree with your comment. This PR is a proof of concept and the result of my inability to do the right thing in the first place (though I tried for a bit). I'll gladly try and update the PR as you suggest, but I might need some guidance (my C/C++ is rusty to say the least, my Python C API knowledge amounts to zero and my experience with OpenGL predates programmable pipelines).

My understanding is that I should change `create_context()` in ModernGL.cpp, near this piece of code:

```c++
    // Map OpenGL functions
    void ** gl_function = (void **)&ctx->gl;
    for (int i = 0; GL_FUNCTIONS[i]; ++i) {
        PyObject * val = PyObject_CallMethod(ctx->ctx, "load", "s", GL_FUNCTIONS[i]);
        if (!val) {
            return NULL;
        }
        gl_function[i] = PyLong_AsVoidPtr(val);
        Py_DECREF(val);
    }
```

If any of the `glProgramUniform*()` returns NULL, I should instead pass a pointer to the wrapper function. My question is where should I store the addresses of `glUniform*()` functions (and `glUseProgram()`) so that said wrapper functions may call them? That can't be mere global variables, and should instead be attached to the context, right? But if so, the wrapper functions should be closures that embed the context's address. Is that at all possible in C? This is clearly going to places beyond my command of C/C++ :)

szabolcsdombi:
reimplemented uniform setters with the proper methods.
https://github.com/moderngl/moderngl/pull/522

PR #477, opened 2021-08-31
Docs: update ci install docs to use env vars

naveen521kk:
### Description

There is support for using environment variables for finding opengl32.dll on glcontext.
Use that in CI docs.

I didn't know about [this](https://github.com/moderngl/glcontext#environment-variables) before :sweat_smile:.


einarf:
Thanks!

PR #479, opened 2021-09-26
C fixups

encukou:
Here are issues I found when trying moderngl with Python 3.10 and 3.11-dev.

### Use Py_SET_TYPE

In Python 3.11a0dev (main branch), Py_TYPE is a function instead
of a macro, which means it cannot be assigned to.
For assignments, Py_SET_TYPE should be used. That was only added
in Python 3.9, but for older versions we can use a simple macro.

It's possible that this change will be reverted before Python 3.11
is released, but still, assigning to Py_TYPE is deprecated.

###  Fix error handling in `MGLError_SetTrace`

(This turned a error in my code into a fatal crash, making it hard to debug.)

When `MGLError_SetTrace` was called with an active exception
(such as from `MGLFramebuffer_set_scissor` after passing a
non-int element), `PyImport_ImportModule` was called with an
exception set.
The effect on the import machinery is unpredictable: generally
the first check of `PyErr_Occurred` (rather than of the C return
value) will assume a function failed.
In Python 3.10 this leads to a fatal crash.

Clear the old exception *first* since we replace it with a new one.
(The super-proper way to do this would be exception chaining,
"During handling of the above exception, another exception occurred".
The current C API for that is complicated, but there are [plans to fix that](https://github.com/python/cpython/pull/27799))

### DIFF
--- a/moderngl/src/Attribute.cpp
+++ b/moderngl/src/Attribute.cpp
@@ -61,7 +61,7 @@ void MGLAttribute_Invalidate(MGLAttribute * attribute) {
 
 	// TODO: decref
 
-	Py_TYPE(attribute) = &MGLInvalidObject_Type;
+	Py_SET_TYPE(attribute, &MGLInvalidObject_Type);
 	Py_DECREF(attribute);
 }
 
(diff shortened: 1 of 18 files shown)


szabolcsdombi:
thank you

Issue #242, opened 2018-08-23
Depth Sampler - can't retrieve depth information from shader


yoyonel:
**Description**

I try to use depth buffer/texture (renderbuffer) in RTT (Render To Target) mode in a shader (for example for shadow rendering).
The depth buffer and texture seems to be rightly fill, but the sampler fetch (inside the shader rendering) always return 0.0 (null value).
I follow your instruction inside texture.py:
```py
            By default depth textures have ``GL_TEXTURE_COMPARE_MODE`` set to
            ``GL_COMPARE_REF_TO_TEXTURE``, meaning any texture lookup will
            return a depth comparison value.

            If you need to read the actual depth value in shaders, setting
            ``compare_func`` to a blank string will set ``GL_TEXTURE_COMPARE_MODE`` to
            ``GL_NONE`` making you able to read the depth texture as a ``sampler2D``::

                uniform sampler2D depth;
                out vec4 fragColor;
                in vec2 uv;

                void main() {
                    float raw_depth_nonlinear = texture(depth, uv);
                    fragColor = vec4(raw_depth_nonlinear);
                }
```
but ... it don't work ...  (always 0.0 return values)

**Proof of code**

```py
"""

"""
import numpy as np
import moderngl
import moderngl as mgl


def main():
    ctx = mgl.create_standalone_context()
    ctx.enable(mgl.DEPTH_TEST)
    prog = ctx.program(
        vertex_shader='''
            #version 330

            in vec3 in_vert;

            void main() {
                gl_Position = vec4(in_vert, 1.0);
            }
        ''',
        fragment_shader='''
            #version 330

            void main() {
                //gl_FragDepth = 0.1234;
            }
        ''',
    )

    prog_for_depth = ctx.program(
        vertex_shader='''
            #version 330

            in vec2 in_vert;
            out vec2 uv;

            void main() {
                uv = in_vert;
                gl_Position = vec4(in_vert * 2.0 - 1.0, 0.0, 1.0);
            }
        ''',
        # from `texture.py`
        fragment_shader='''
            #version 330

            uniform sampler2D depth;
            out vec4 fragColor;
            in vec2 uv;

            void main() {
                float raw_depth_nonlinear = texture(depth, uv).r;
                fragColor = vec4(raw_depth_nonlinear);
            }
        ''',
    )

    prog_for_depth['depth'].value = 0

    vertices = np.array([
        [-1, 1, 0.],
        [-1, -1, 0.],
[...]

Issue #161, opened 2018-01-25
Shadowmapping example


jakebruce:
Hi there, thanks for your work on this very promising wrapper.

It would be nice to have a shadow mapping example. I've been trying to implement basic shadow mapping according to a [tutorial](http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-16-shadow-mapping/) but I'm unsure how to do some of it (and whether it's even implemented) in the ModernGL API.

For example, according to [this stackoverflow post](https://stackoverflow.com/questions/22419682/glsl-sampler2dshadow-and-shadow2d-clarification) it is import to set `GL_TEXTURE_COMPARE_MODE` = `GL_COMPARE_REF_TO_TEXTURE​`. Is this implemented in ModernGL?

I'm also wondering about depth textures as render targets (without a color attachment). Is this possible in ModernGL? Do we have to use a framebuffer with a dummy color attachment that we then ignore?

Cheers!

cprogrammer1994:
Hi! Great catch! I will add the missing `GL_TEXTURE_COMPARE_MODE` for depth textures. depth_textures will have a setter for `GL_TEXTURE_COMPARE_FUNC` too.

- [glTexParameter](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexParameter.xhtml)

cprogrammer1994:
> I'm also wondering about depth textures as render targets (without a color attachment). Is this possible in ModernGL? Do we have to use a framebuffer with a dummy color attachment that we then ignore?

Currently framebuffers must have at least 1 color attachment. I will investigate what will happen if I remove that check.

cprogrammer1994:
In ModernGL5 Framebuffers can have 0 color_attachments if a depth_attachment was provided.
I will add an example soon.

Textures have a getsetter for `compare_func` (available if the texture is a depth texture)
The `GL_TEXTURE_COMPARE_MODE` is set by default.

PR #480, opened 2021-11-02
Examples: rendering scene-1 (terrain with textured billboard) with shadows

yoyonel:
### Description

Implement #161
Related to #242

Using scene model from `examples/data/scene-1.obj` like the existing example ``:
image
with shadows:
image

From the previous example, i'm keeping the lighting diffuse colored and textured rendering pass and add a very simple shadows rendering with a Shadow Mapping (hardware Percentage Close Filtering 2x2) technique.
Some references (not exhaustive):
- http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-16-shadow-mapping
- https://community.khronos.org/t/slope-scale-depth-bias-in-opengl-3-2-core/62194/9
- https://docs.microsoft.com/en-us/windows/win32/dxtecharts/common-techniques-to-improve-shadow-depth-maps?redirectedfrom=MSDN

I hope you're like it :smiley: 

### List of changes

- added an example `examples/shadowmapping.py`

### DIFF
--- a/examples/shadowmapping.py
+++ b/examples/shadowmapping.py
@@ -1,0 +1,242 @@
+"""
+Links:
+- http://www.opengl-tutorial.org/fr/intermediate-tutorials/tutorial-16-shadow-mapping/#shadowmap-basique
+- https://learnopengl.com/Advanced-Lighting/Shadows/Shadow-Mapping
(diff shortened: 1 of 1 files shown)


einarf:
Thanks!

PR #481, opened 2021-11-03
Examples: Advanced particle system

yoyonel:
### Description

Extending the `example/particle_system.py` by adding a transform vertex (buffer) pass on points sprites.
Using a gold noise in shaders (from shadertoy example) and animated the particle system fully in GPU.

Hope you're like it :smile: 

At start:
image
loop on animation:
image

### List of changes

- added an example `example/particle_system_advanced.py`


einarf:
Thanks again!

PR #483, opened 2021-11-08
fix segmentation fault when reading the provoking vertex

szabolcsdombi:


einarf:
Nice. I never spotted that one.

Issue #472, opened 2021-07-19
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install moderngl’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPi.

Resolution
-----------
On aarch64, ‘pip install moderngl’ should download the wheels from PyPi

@einarf, Please let me know your interest in releasing aarch64 wheels. I can help with this.

einarf:
We'll move to github actions instead of travis and can probably include it there. We can keep this one open until then.

einarf:
There should be binaries on PyPI now

PR #488, opened 2021-12-16
Add linux aarch64 wheel build support

odidev:
Added aarch64 wheel build support.

Related to #472, @einarf Could you please review this PR?


einarf:
This should already be resolved. Thanks a lot of the RP!

PR #494, opened 2022-02-25
docs: update installation docs related to mesa

naveen521kk:
### Description

There is support for using environment variables for finding opengl32.dll on glcontext.
I made this change for CI docs in https://github.com/moderngl/moderngl/pull/477 but
forgot to update the same here.

### List of changes

- **added**: Info about using environment variable for opengl32.dll in mesa install docs
- **fixed**: Some formatting changes


einarf:
Thanks!

Issue #447, opened 2021-03-05
Failing to read multisampled framebuffers


einarf:
Apparently this is a bug that was never reported here.
https://stackoverflow.com/questions/61302400/how-to-enable-anti-aliasing-in-moderngl-egl-backend

I see a couple of projects using pyopengl in addition to moderngl to solve this. That again causes other issues.

If not a bug it's at least something several people are struggling with and we might need to provide a solution.

szabolcsdombi:
ms textures need to be downsampled before reading. ctx.copy_framebuffers is the method that can do it. I am sure we used this extensively in the past.

I have to add an example before this one gets closed.

szabolcsdombi:
https://github.com/szabolcsdombi/headless-moderngl-experiment/blob/master/src/main_multisample.py#L68

maciejhalber:
It is possible that I should open a new issue for this, however, there currently does not seem to be a way to perform this operation with multiple color attachments.

Looking at the code here: https://github.com/moderngl/moderngl/blob/master/moderngl/src/Context.cpp#L25, when using the glBlitFramebuffer, the targets are not set using glReadBuffer / glDrawBuffer calls (https://www.khronos.org/opengl/wiki/Framebuffer#Read_color_buffer)

Here is a use case that does not work:
```
FBO_A - multisampled, two color attachment
FBO_B - no multisampling, two color attachments

copy_framebuffer(FBO_B, FBO_A)

FBO_B.read(components=4, attachment=1) --> This returns expected image
FBO_B.read(components=4, attachment=1) --> This returns all zeros.
```

szabolcsdombi:
Do you mean

```
FBO_B.read(components=4, attachment=1) --> This returns expected image
FBO_A.read(components=4, attachment=1) --> This returns all zeros.
```

?

maciejhalber:
Ah, I am very sorry - it was a copy paste error on my part. The corrected version of this pseudo code is:
```
FBO_A - multisampled, two color attachment
FBO_B - no multisampling, two color attachments

copy_framebuffer(FBO_B, FBO_A)

FBO_B.read(components=4, attachment=0) --> This returns expected image
FBO_B.read(components=4, attachment=1) --> This returns all zeros.
```

I also am attaching a full python script that I would expect to show 3 images - red, blue and green triangle. 
If you run it with moderngl version 5.6.4, it shows picture of a red triangle and two empty images.
```
import moderngl
import numpy as np
from PIL import Image
[...]

PR #513, opened 2022-08-01
Copy all multisampled color attachments

maciejhalber:
### Description

This PR is modifying the way `Context.copy_framebuffer` method operates when copying from mutisampled FBO with multiple color attachments into a non-multisampled FBO with multiple color attachment. 

Such functionality is needed when user wants to read out multiple color attachments from a multisampled FBO to the CPU memory.

Previously, only the first color attachments (`GL_COLOR_ATTACHMENT0`) was copied, while others have been ignored.

This PR addresses the [issue #447](https://github.com/moderngl/moderngl/issues/447) 

### List of changes

- **changed** `moderngl/src/Context.cpp`. Code first checks whether the source and destination buffers have the same number of color attachments. If so, the current state of `GL_READ_BUFFER` and `GL_DRAW_BUFFER` is recorded. Next, the code loops over the `draw_buffers` member of `MGLFramebuffer` objects, and uses `glReadBuffer` \ `glDrawBuffer` calls to set the targets for `glBlitFramebuffer` operation. Finally, the code restores the previous state of `GL_READ_BUFFER` and `GL_DRAW_BUFFER` in order to avoid any side effects.

### Example program:
On the main branch, this program shows one image of red triangle and two empty images.
With the propsed changes, the programs shows shows images of red, green and blue triangles, as expected.
```py
import moderngl
import numpy as np
from PIL import Image
from PIL.Image import Transpose


def multisampled_mrt():
    with moderngl.create_context(
        require=330,
        standalone=True,
        backend="egl",
    ) as ctx:
        # Init the shader program
        prog = ctx.program(
            vertex_shader="""
                #version 330
                in vec2 in_vert;

                void main() {
                    gl_Position = vec4(in_vert, 0.0, 1.0);
                }
            """,
            fragment_shader="""
                #version 330
                
                in vec3 v_color;
                
                layout(location=0) out vec4 f_color_r;
                layout(location=1) out vec4 f_color_g;
                layout(location=2) out vec4 f_color_b;

                void main() {
                    f_color_r = vec4(1.0, 0.0, 0.0, 1.0);
                    f_color_g = vec4(0.0, 1.0, 0.0, 1.0);
                    f_color_b = vec4(0.0, 0.0, 1.0, 1.0);
                }
            """,
        )
        resolution = (800, 600)

        # Create multisampled framebuffer with multiple color attachments
        color_r_texture = ctx.texture(resolution, components=4, dtype="f1", samples=4)
        color_g_texture = ctx.texture(resolution, components=4, dtype="f1", samples=4)
        color_b_texture = ctx.texture(resolution, components=4, dtype="f1", samples=4)

        fbo = ctx.framebuffer(
            (color_r_texture, color_g_texture, color_b_texture),
            ctx.depth_renderbuffer(resolution, samples=4),
        )

        # Create VAO
        vertices = np.array(
            [0.0, 0.8, -0.6, -0.8, 0.6, -0.8],  # x, y
            dtype="f4",
        )

        vbo = ctx.buffer(vertices)

        vao = ctx.vertex_array(prog, [(vbo, "2f", "in_vert")])

        # Render to offscreen, multisampled framebuffer

        fbo.use()
        fbo.clear(color=(0.0, 0.0, 0.0, 1.0))
        vao.render()

        # Create "Resolve" framebuffer to which we copy multisampled framebuffer
        color_r_resolve = ctx.texture(resolution, components=4, dtype="f1", samples=0)
        color_g_resolve = ctx.texture(resolution, components=4, dtype="f1", samples=0)
        color_b_resolve = ctx.texture(resolution, components=4, dtype="f1", samples=0)
        resolve_fbo = ctx.framebuffer(
            (color_r_resolve, color_g_resolve, color_b_resolve),
            ctx.depth_renderbuffer(resolution, samples=0),
        )

        # Perform the copy
        ctx.copy_framebuffer(resolve_fbo, fbo)

        # Read images out
        img_red = np.frombuffer(
            resolve_fbo.read(components=4, attachment=0), dtype=np.uint8
        ).reshape((resolution[1], resolution[0], 4))
        img_green = np.frombuffer(
            resolve_fbo.read(components=4, attachment=1), dtype=np.uint8
        ).reshape((resolution[1], resolution[0], 4))
        img_blue = np.frombuffer(
            resolve_fbo.read(components=4, attachment=2), dtype=np.uint8
        ).reshape((resolution[1], resolution[0], 4))

        # Display images
        Image.fromarray(img_red).transpose(Transpose.FLIP_TOP_BOTTOM).show()
        Image.fromarray(img_green).transpose(Transpose.FLIP_TOP_BOTTOM).show()
        Image.fromarray(img_blue).transpose(Transpose.FLIP_TOP_BOTTOM).show()


if __name__ == "__main__":
    multisampled_mrt()
```

### DIFF
--- a/moderngl/src/Context.cpp
+++ b/moderngl/src/Context.cpp
@@ -288,6 +288,7 @@ PyObject * MGLContext_copy_framebuffer(MGLContext * self, PyObject * args) {
 		{
 			gl.ReadBuffer(src->draw_buffers[i]);
 			gl.DrawBuffer(dst_framebuffer->draw_buffers[i]);
+
 			gl.BlitFramebuffer(
 				0, 0, width, height,
 				0, 0, width, height,
@@ -298,6 +299,7 @@ PyObject * MGLContext_copy_framebuffer(MGLContext * self, PyObject * args) {
 		gl.BindFramebuffer(GL_FRAMEBUFFER, self->bound_framebuffer->framebuffer_obj);
 		gl.ReadBuffer(prev_read_buffer);
 		gl.DrawBuffer(prev_draw_buffer);
+		gl.DrawBuffers(self->bound_framebuffer->draw_buffers_len, self->bound_framebuffer->draw_buffers);
 
 	} else if (Py_TYPE(dst) == &MGLTexture_Type) {
 
--- a/moderngl/src/Context.cpp
+++ b/moderngl/src/Context.cpp
@@ -269,15 +269,35 @@ PyObject * MGLContext_copy_framebuffer(MGLContext * self, PyObject * args) {
 			height = src->height < dst_framebuffer->height ? src->height : dst_framebuffer->height;
 		}
 
-		gl.BindFramebuffer(GL_READ_FRAMEBUFFER, src->framebuffer_obj);
-		gl.BindFramebuffer(GL_DRAW_FRAMEBUFFER, dst_framebuffer->framebuffer_obj);
-		gl.BlitFramebuffer(
-			0, 0, width, height,
-			0, 0, width, height,
-			GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT,
-			GL_NEAREST
-		);
+		if (dst_framebuffer->draw_buffers_len != src->draw_buffers_len)
+		{
+			MGLError_Set("Destination and source framebuffers have different number of color attachments!");
+			return 0;
+		}
+
+		int prev_read_buffer = -1;
+		int prev_draw_buffer = -1;
+		gl.GetIntegerv(GL_READ_BUFFER, &prev_read_buffer);
+		gl.GetIntegerv(GL_DRAW_BUFFER, &prev_draw_buffer);
+
+		int color_attachment_len = dst_framebuffer->draw_buffers_len;
+		for (int i = 0; i < color_attachment_len; ++i)
+		{
+			gl.ReadBuffer(src->draw_buffers[i]);
+			gl.DrawBuffer(dst_framebuffer->draw_buffers[i]);
+
+			gl.BindFramebuffer(GL_READ_FRAMEBUFFER, src->framebuffer_obj);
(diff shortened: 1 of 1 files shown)


moderngl/src/Context.cpp
```diff
@@ -269,15 +269,35 @@ PyObject * MGLContext_copy_framebuffer(MGLContext * self, PyObject * args) {
...
+			gl.BlitFramebuffer(
+				0, 0, width, height,
+				0, 0, width, height,
+				GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT,
+				GL_NEAREST
+			);
+		}
 		gl.BindFramebuffer(GL_FRAMEBUFFER, self->bound_framebuffer->framebuffer_obj);
+		gl.ReadBuffer(prev_read_buffer);
+		gl.DrawBuffer(prev_draw_buffer);
```
szabolcsdombi, line 300: It has to be checked if this messes up theself->bound_framebuffer's `glDrawBuffers` call (please note the s at the end)
I assume gl.DrawBuffer(x) is equivalent of `glDrawBuffers(1, &{x})` so setting all the remaining attachments to GL_NONE.

Maybe we want https://github.com/moderngl/moderngl/blob/master/moderngl/src/Framebuffer.cpp#L405 instead

maciejhalber: That is an excellent point! 

That being said, OpenGL docs are not completely clear on the differences between glDrawBuffer and glDrawBuffers, esp. in the context of glBlitFramebuffer operation. I thought I could replace my calls to glDrawBuffers, but that results in an odd behaviour where all images show blue triangle. So it seems (empirically) that separate glDrawBuffers are needed for proper copying (?). 

I defensively added call to glDrawBuffers to restore the state stored in the `bound_framebuffer`. Let me know what you think.

maciejhalber:
Hi again! I wanted to check back in to see if there are any other concerns with this PR that I could address? 
Thanks!

PR #515, opened 2022-08-17
Remove redundant code from example

kmaork:
Removed two functions that are the same as the functions they override


einarf:
Thanks!

einarf:
Why remove the initializer?

kmaork:
Hey, I removed it because it's redundant - it only calls super with the same arguments. Removing it will have no effect except for clearing up the code a little bit :)

einarf:
Yup. Derp. Let's just merge this.

PR #516, opened 2022-10-08
Fix a few typos

Starbuck5:
### Description

I was looking at the ModernGL docs and noticed it said "byte rage" instead of "byte range." I thought that if I was going to make a PR anyway, I might as well see if I can find other typos, so I used `codespell` to span the code for typos.

`codespell` reports a few more typos in actual source files, but I wanted this to be a simple PR so I didn't want to touch any actual source code.

### List of changes

- **edited** docs/ and examples/ to fix typos

### DIFF
--- a/docs/install/installation.rst
+++ b/docs/install/installation.rst
@@ -61,8 +61,8 @@ this method, to make Moderngl work.
 
 There are essentially two ways,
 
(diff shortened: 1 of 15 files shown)


einarf:
Thanks!

PR #521, opened 2022-10-22
Added support for custom internal formats to cubemaps

yonibr:


moderngl/context.py
```diff
@@ -1307,6 +1307,7 @@ def texture_cube(
         *,
         alignment: int = 1,
         dtype: str = 'f1',
+        internal_format: int = None,
```
einarf, line 1310: Need docstring for this new keyword argument. Can probably just copy from `ctx.texture()`

einarf: Could we just change the typing to `Optional[int]` for both cube and texture? The current typing is not correct.

einarf:
`TextureCube` was fixed in : https://github.com/moderngl/moderngl/pull/521

The remaining texture types should also be covered here

PR #523, opened 2022-10-24
Stability improvements

szabolcsdombi:


yoyonel:
Hi,
Why removing the (cpp) files hierarchy and put the source code in a single `modern.cpp` file ?

szabolcsdombi:
- the moderngl core should be only responsible for the low level stuff, the final code should be < 5k loc
- one file compiles way faster, and for bug reporters / pull requesters it is way easier to browse/patch one file.
- the current hierarchy is redundant in function defs and similar. it at least takes 3 places to touch to add a method like the create_context.

As a proof this is a good idea: most of moderngl is implemented in zengl in < 3k loc ([here](https://github.com/szabolcsdombi/zengl/blob/main/zengl.cpp))
high level stuff like validation of an entire pipeline is moved to Python ([here](https://github.com/szabolcsdombi/zengl/blob/main/_zengl.py#L434-L546))
actually moderngl itself have some nice new prototypes in the repo labelled as moderngl-next but for non backward compatible reasons they never became the next major version for moderngl. no such prototype will.

When a moderngl issue is reported, I often find it easier to do the fix myself, than to explain how many files to touch at who knows how many places to get it done.

The less files you have to open, the less lines you have to read the more likely you won't give up on fixing a bug.

With learned helplessness taken into account one might not even try to figure out what to touch when the number of files to choose from is high.

---

Anyways this branch was mostly a proof of contept for me to find out if such a refactor on moderngl is possible.

The results are positive and with smaller steps it would make sense.
But first some parts must be fixed before it can happen.

yoyonel:
> * the moderngl core should be only responsible for the low level stuff, the final code should be < 5k loc
> * one file compiles way faster, and for bug reporters / pull requesters it is way easier to browse/patch one file.
> * the current hierarchy is redundant in function defs and similar. it at least takes 3 places to touch to add a method like the create_context.
> 
> As a proof this is a good idea: most of moderngl is implemented in zengl in < 3k loc ([here](https://github.com/szabolcsdombi/zengl/blob/main/zengl.cpp)) high level stuff like validation of an entire pipeline is moved to Python ([here](https://github.com/szabolcsdombi/zengl/blob/main/_zengl.py#L434-L546)) actually moderngl itself have some nice new prototypes in the repo labelled as moderngl-next but for non backward compatible reasons they never became the next major version for moderngl. no such prototype will.
> 
> When a moderngl issue is reported, I often find it easier to do the fix myself, than to explain how many files to touch at who knows how many places to get it done.
> 
> The less files you have to open, the less lines you have to read the more likely you won't give up on fixing a bug.
> 
> With learned helplessness taken into account one might not even try to figure out what to touch when the number of files to choose from is high.
> 
> Anyways this branch was mostly a proof of contept for me to find out if such a refactor on moderngl is possible.
> 
> The results are positive and with smaller steps it would make sense. But first some parts must be fixed before it can happen.

Thx for this awsome/detailled answer, it's very interesting ^^
If you need some feedback and/or help on the python side, i'm here :)

PR #536, opened 2022-12-12
Revamp tests

einarf:
Rewrote all tests to use pytest over python unittest. They used a mix of both in the past.

Tests are now a lot more strict checking context for errors in setup and teardown between each test to easily locate issues. Most of the tests should be a lot simpler now. We can look into consolidating tests modules and add missing tests in the future. This pass is simply converting each tests module one by one and locating issues. Having more shared fixtures is also something that can be done, but this might also increase the time needed to understand some of the tests. Right now it should be fairly straight forward.

This also includes a few bug fixes encountered during the rewrite.
* Fixed a regresson causing objects being released twice to crash
* Attribute didn't have the location value set
* Don't attempt to set anisotropy in samplers when not supported
* Samplers should have 0 anisotropy by default

TODO before merge:
* [x] Runs with llvmpipe in WSL2
* [x] Runs on linux with llvmpipe
* [x] Runs on linux with Nividia/AMD
* [x] Runs on windows Nvidia
* [x] Runs on windows AMD
* [x] Runs on MacOS

Issues
* [x] Framebuffer mask issues on windows/amd (driver issue)
* [x] Scissor tests failing on AMD/Linux (due to random data in attachment)


szabolcsdombi:
solved, the array size was zero. it was not deduced by the compiler

einarf:
ah right. I guess that only really works in geo shader inputs then.

Issue #528, opened 2022-11-27
Expose the internal glcontext context, Also accept one in the create_context


szabolcsdombi:
It should be possible to access the glcontext context created by moderngl.

maybe `ctx.mglo.ctx`

It should be possible to pass a glcontext context to the moderngl create_context.

```py
backend = glcontext.default_backend()
gl_ctx = backend()
ctx = moderngl.create_context(context=gl_ctx)
```

szabolcsdombi:
https://github.com/moderngl/moderngl/issues/527

PR #538, opened 2022-12-14
accept context as a parameter

szabolcsdombi:


szabolcsdombi:
Resolves https://github.com/moderngl/moderngl/issues/528

Issue #540, opened 2022-12-20
Compute Shader: Introspect blocks shader storage blocks


einarf:
Compute shaders introspect uniforms and uniforms blocks, but not shader storage blocks.

einarf:
Got something quick working locally
```py
import moderngl
import OpenGL.GL as gl

ctx = moderngl.create_context(standalone=True)

source = '''
    #version 430

    layout(local_size_x=1, local_size_y=1, local_size_z=1) in;

    uniform float scale;

    layout (std430, binding = 3) readonly buffer in0
    {
        vec4 inBuffer[];
    };

    layout (std430, binding = 2) writeonly buffer out0
    {
        vec4 outBuffer[];
    };

    void main(void)
    {
        uint index = gl_GlobalInvocationID.x;
        outBuffer[index] = inBuffer[index] * 2.0 * scale;
    }
'''

print(cs._members)
```
out 
```
{
    'scale': <Uniform: 0>,
    'in0': <StorageBlock: 0>,
    'out0': <StorageBlock: 1>
}
```

einarf:
The question is. Should we maybe unify this because Program can also have storage blocks?

dibenedetto:
> The question is. Should we maybe unify this because Program can also have storage blocks?

Please remember to report correct binding locations (in this case 3 and 2, respectively).

einarf:
Better to comment in the issue. I will probably merge it today.

dibenedetto:
> Better to comment in the issue. I will probably merge it today.

I commented in the issue. What do you mean? PR?

einarf:
yeah the PR, sorry 😄

einarf:
Ignore me. I'm doing too many things at the same time here 😆

einarf:
Find storage block members
```py
cs = ctx.compute_shader(source)
for name in cs:
    obj = cs[name]
    if isinstance(obj, moderngl.StorageBlock):
        print(f"{obj.name}: {obj} binding={obj.binding}")
```
Output
```
in0: <StorageBlock: 0> binding=3
out0: <StorageBlock: 1> binding=2
```

Setting binding
```py
in_0 = cs['in0']
out_0 = cs['out0']
print(in_0.binding)
print(out_0.binding)
in_0.binding = 4
out_0.binding = 5
print(in_0.binding)
print(out_0.binding)
```
Output
```
3
2
4
5
```

dibenedetto:
Great!

einarf:
[...]

PR #541, opened 2022-12-20
Introspect storage blocks #540

einarf:
Quick and dirty introspection of storage blocks. It works, but needs to look through it.

I do wonder if we should move program introspection to a function in the context so we can reuse code between program and compute shader. Just need to check version code and skip the ssbo part?

### DIFF
--- a/_moderngl.py
+++ b/_moderngl.py
@@ -118,6 +118,38 @@ class UniformBlock:
         self.ctx._set_ubo_binding(self.program_obj, self.index, value)
 
 
+class StorageBlock:
+    def __init__(self):
+        self.program_obj = None
+        self.index = None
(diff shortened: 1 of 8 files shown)


szabolcsdombi:
the reason this was not implemented is that storage blocks are only available in gl versions where explicit binding is also available.
it is good to have this implemented as well.

szabolcsdombi:
maybe we also need to update the .pyi file.

einarf:
> maybe we also need to update the .pyi file.

Definitely. I still need to look through this at least one more time. It was a quick start.

einarf:
That should be all. Need to clean up some docstrings etc, but that can be done separately.

PR #544, opened 2023-01-08
docs: Change "do" to "does" in intro header

unparalleled-js:
### Description

Correct me if I am wrong, but would "does" make more sense here?
I am following the tutorials but thought I'd fix this as I went.

### List of changes

- **added** ...
- **removed** ...
- **fixed** ...
- **changed** ...


einarf:
Thanks. I just extended extended the header line by two more characters

PR #558, opened 2023-02-23
intregration_pycairo.py: remove geometry import

gregorni:
### Description

In examples/integration_pycairo.py, there was an import which didn't seem to be used.

### List of changes

- **removed** the import of moderngl_window.geometry


einarf:
Thanks!

Issue #467, opened 2021-06-03
Rendering with empty framebuffers


0x23:
In opengl there is the possibility to render with framebuffers that have no images attached. The following is a cite from [https://www.khronos.org/opengl/wiki/Framebuffer_Object](https://www.khronos.org/opengl/wiki/Framebuffer_Object) that describes the usecase quite well:

> It is possible to render to a framebuffer object that has no attachments. Obviously none of the fragment shader outputs will be written to anywhere in this case, but rendering can otherwise proceed as normal. This is useful for using arbitrary reading and writing of image data from shaders, rather than writing to a bound framebuffer.

In this case we would need a framebuffer creation method in moderngl that justs takes the necessary (four) framebufer parameter but no color/depth atachments:

Internally those can be parameter can be set using `void glFramebufferParameteri(GLenum target​, GLenum pname​, GLint param​);`
wich supports:
 - GL_FRAMEBUFFER_DEFAULT_WIDTH
 - GL_FRAMEBUFFER_DEFAULT_HEIGHT
 - GL_FRAMEBUFFER_DEFAULT_LAYERS 
 - GL_FRAMEBUFFER_DEFAULT_FIXED_SAMPLE_LOCATIONS 

**SUMMARY:**
a new function: `create_empty_framebuffer(width, height, layers=0, fixed_sample_locations=0)` that uses `glFramebufferParameteri` internally to set the given parameter and creates a framebuffer object without attachments.

einarf:
> This is useful for using arbitrary reading and writing of image data from shaders, rather than writing to a bound framebuffer.

Do you have more specific example? This looks easy enough. I just want to have one clear usecase with very specific details that can be used as a unit test.

0x23:
My use case is to write each rendered fragment (not only the closest one) to a shader storage buffer to implement order independent transparency. As all data is soterd in this shader storage buffer (wich i can index arbitrarily) i dont need the actual framebuffer.
In my case I deal with quite large images (20-50Mpixel) so the momory savings of an empty framebuffer would be very welcome.
I am not shure if this is a good usecase for a unit test though.

The simplest example I can think of is to count the total number of rendered fragments in a scene by using atomicAdd in the fragment shader to increment a single global counter - no framebuffer needed.

einarf:
ah right. so you just want the output in a shader storage buffer 👍 

You could also mix in pyopengl for this meanwhile. If you do, please be free to share snippets

0x23:
[...]

PR #564, opened 2023-02-25
Render with empty framebuffers

szabolcsdombi:
#467


mabl:
Thank you :-) I'll give this a try as soon as possible!

szabolcsdombi:
there is an example in the commit. I hope it helps.

mabl:
That worked perfectly. The addition of memory barriers is also very welcome. Thanks again.

PR #574, opened 2023-03-10
Fixed ComputeShader.run_indirect() and added missing method in cpp ob…

ramenguy99:
### Description

Indirect dispatch in compute shader added in #563 curently does not work as it's invoking the `run` method, and the `run_indirect` method does not exist.

### List of changes

- Fixed `run_indirect()` method in `ComputeShader`.
- Added missing method in compute shader cpp implementation.

### DIFF
--- a/moderngl/__init__.py
+++ b/moderngl/__init__.py
@@ -250,7 +250,7 @@ class ComputeShader:
         return self.mglo.run(group_x, group_y, group_z)
 
     def run_indirect(self, buffer: 'Buffer', offset: int = 0) -> None:
-        return self.mglo.run(buffer, offset)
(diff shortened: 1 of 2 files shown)


szabolcsdombi:
thank you

ramenguy99:
Hey, thanks for merging @szabolcsdombi. Is there a plan for a 5.8.2 release that includes this change? It would be helpful to be able to use this in packages distributed with pip.

szabolcsdombi:
Hi,

I tagged [5.8.2](https://github.com/moderngl/moderngl/releases/tag/5.8.2) and soon it will be available on PyPI.

Issue #578, opened 2023-04-15
Add depth cubemap functionality


yonibr:
Example use cases:
- point light shadow maps
- light probes

klattimer:
I'm just trying to figure out how to implement this and hitting "invalid depth attachment" - so there's clearly something missing.

klattimer:
After some digging, I think the problem with the ctx.framebuffer(color_attachment, depth_attachment) call is that TextureCube doesn't have a width/height it has a size - which reflects the size of an individual cube face, but not a width/height which is compared to the size of the color attachment. 

I'm not able to debug the C side unfortunately, to determine if this is actually the case. I've tried using a different size for the color attachment but that hasn't worked in 3x2, 2x3, 6x1, 1x6 arrangements. 

I'm reasonably convinced this is the cause, but I'm unsure how to go about fixing it.

@yonibr would you be able to help?

klattimer:
I tried adding width, height and as it turns out sampler properties however I still get the same problem with an invalid depth texture.

yonibr:
Hi @klattimer, when I implemented depth cubemaps, I didn't also modify moderngl's framebuffers to support depth cubemaps as the depth attachment, so I think you'll have to either implement that yourself or manually create the framebuffer using pyopengl. Here's how I do it:
```
# Create the shadow map depth texture
self.shadow_map = ctx.depth_texture_cube((self.width, self.height))
self.shadow_map.filter = moderngl.LINEAR, moderngl.LINEAR
self.shadow_map.compare_func = ''

self.shadow_map_alpha = ctx.texture_cube((self.width, self.height), 1)
self.shadow_map.filter = moderngl.LINEAR, moderngl.LINEAR

# Create the framebuffer object
depth_map_fbo = self.depth_map_fbo = glGenFramebuffers(1)
glBindFramebuffer(GL_FRAMEBUFFER, depth_map_fbo)
glFramebufferTexture(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, self.shadow_map.glo, 0)
glFramebufferTexture(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, self.shadow_map_alpha.glo, 0)
glDrawBuffer(GL_COLOR_ATTACHMENT0)
glBindFramebuffer(GL_FRAMEBUFFER, 0)
```

klattimer:
I've done a bit of digging, I'd rather upstream a patch to fix it for everyone than work around it. I might have something later today.

PR #579, opened 2023-04-15
Implemented depth cube maps

yonibr:
See https://github.com/moderngl/moderngl/issues/578

### DIFF
--- a/src/moderngl.cpp
+++ b/src/moderngl.cpp
@@ -273,11 +273,12 @@ struct MGLTextureCube {
     int texture_obj;
     int width;
     int height;
-    int depth;
+    bool depth;
     int components;
     int min_filter;
     int mag_filter;
     int max_level;
+	int compare_func;
     float anisotropy;
     bool released;
 };
@@ -6939,6 +6940,8 @@ PyObject * MGLContext_texture_cube(MGLContext * self, PyObject * args) {
     texture->components = components;
     texture->data_type = data_type;
 
+    texture->depth = false;
+
     texture->min_filter = data_type->float_type ? GL_LINEAR : GL_NEAREST;
     texture->mag_filter = data_type->float_type ? GL_LINEAR : GL_NEAREST;
     texture->max_level = 0;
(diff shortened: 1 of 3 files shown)


src/moderngl.cpp
```diff
@@ -273,11 +273,12 @@ struct MGLTextureCube {
     int texture_obj;
     int width;
     int height;
-    int depth;
+    bool depth;
```
szabolcsdombi, line 276: int is preferred for booleans for better C compatibility, mostly due to match the return type of the used Python C Api like the PyObject_IsTrue or the PyArg_ParseTuple.. "p" format.

below released was defined as bool as it is some temporary workaround for a not yet solved problem.

src/moderngl.cpp
```diff
@@ -273,11 +273,12 @@ struct MGLTextureCube {
...
     int components;
     int min_filter;
     int mag_filter;
     int max_level;
+	int compare_func;
```
szabolcsdombi, line 281: please replace tabs with 4 spaces.
with vscode just press `ctrl`+`shift`+`p` then type `>Convert Indentation to Spaces` should do the trick

szabolcsdombi:
looks good to me

src/moderngl.cpp
```diff
@@ -7530,10 +7530,10 @@ PyObject * MGLTextureCube_get_swizzle(MGLTextureCube * self, void * closure) {
 int MGLTextureCube_set_swizzle(MGLTextureCube * self, PyObject * value, void * closure) {
     const char * swizzle = PyUnicode_AsUTF8(value);
 
-	if (self->depth) {
-		MGLError_Set("cannot get swizzle of depth textures");
-		return 0;
-	}
+    if (self->depth) {
+        MGLError_Set("cannot get swizzle of depth textures");
+        return 0;
```
szabolcsdombi, line 7535: errors in a setter must return -1 (https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef)

szabolcsdombi:
thank you very much for implementing this

szabolcsdombi:
i do a bit of testing and maybe add an example too. after that I make a release with this. if I forget it somehow do not hesitate to ping me. until then you can install directly from the master branch with `pip install https://github.com/moderngl/moderngl/archive/refs/heads/master.zip`

Issue #592, opened 2023-07-11
Issues when creating VAOs with SPIR-V shaders.


compik710:
**Description**

I tried to use SPIR-V shaders in practice, but I ran into a problem when creating `ctx.vertex_array()`

What do I see in the console:
```
$ python3 win_test10.py
Traceback (most recent call last):
  File "/home/user_name/win_spirv/win_test10.py", line 31, in <module>
    vao = ctx.vertex_array(prog, buf, 'in_vert', 'in_color')
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/.local/lib/python3.11/site-packages/moderngl/__init__.py", line 1861, in vertex_array
    return self.simple_vertex_array(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/.local/lib/python3.11/site-packages/moderngl/__init__.py", line 1911, in simple_vertex_array
    content = [(buffer, detect_format(program, attributes)) + attributes]
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/.local/lib/python3.11/site-packages/moderngl/__init__.py", line 2243, in detect_format
    return ' '.join('%d%s' % fmt(program[a]) for a in attributes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/.local/lib/python3.11/site-packages/moderngl/__init__.py", line 2243, in <genexpr>
    return ' '.join('%d%s' % fmt(program[a]) for a in attributes)
                                 ~~~~~~~^^^
  File "/home/user_name/.local/lib/python3.11/site-packages/moderngl/__init__.py", line 441, in __getitem__
    return self._members[key]
           ~~~~~~~~~~~~~^^^^^
KeyError: 'in_vert'
```

**Proof of code**

Vertex shader (win.vert):
```glsl
#version 430
layout(location=0) in vec3 in_vert;
layout(location=1) in vec4 in_color;

layout(location=0) out vec4 o_color;
layout(location=1) out float distance;
    
void main() {
    gl_Position = vec4(in_vert.x, in_vert.y, in_vert.z, in_vert.z);
    o_color = in_color;
    distance = in_vert.z;
}
```
[...]

PR #593, opened 2023-07-16
Adding a SPIR-V parser

compik710:
### Description

SPIR-V Level 2 support

The parser is needed to extract the types of input variables from the byte content of SPIR-V. With insignificant code changes, the parser can also extract the types of output variables and variable names (if the compiler left them).

Issue: #592

### List of changes

- **added** parser, parser test and encoding designations in _moderngl.ATTRIBUTE_LOOKUP_TABLE


compik710:
The checks failed, I'll see what I can do.

compik710:
Oh, no, it looks like your repository has strong typing... I'm afraid I'll have to redo the whole parser.

_moderngl.py
```diff
@@ -366,5 +366,168 @@ def make_varying(name, number, array_length, dimension):
...
+            allowed_types[type_id][1] = -1
+        return typ
+
+    for type_id, config in allowed_types.items():
+        assembly(type_id)
+
+    for ids, pointer_type_id in pointer_variable.items():
+        try:
+            extracted_types[ids] = allowed_types[pointer_allowed_types[pointer_type_id]][0]
+        except KeyError:
```
szabolcsdombi, line 448: when do you get a key error?

_moderngl.py
```diff
@@ -366,5 +366,168 @@ def make_varying(name, number, array_length, dimension):
...
+            to_add['class'] = extracted_storage_class_id[ids]
+        if ids in extracted_types:
+            to_add['type'] = extracted_types[ids]
+        if ids in extracted_location:
+            to_add['location'] = extracted_location[ids]
+        extracted_collected[ids]=to_add
+    ##################
+
+    # == Conversion to the moderngl type == #
+    mgl_attr_table = {  # py_type : mgl_type
```
szabolcsdombi, line 474: move this out of the function just like `ATTRIBUTE_LOOKUP_TABLE`

szabolcsdombi:
This can be done without building strings:

```
'i4': 0x1404,  # GL_INT
'i4 vec2': 0x8b53,  # GL_INT_VEC2
...
'f8 vec3 mat4': 0x8f4e,  # GL_DOUBLE_MAT4x3
'f8 vec4 mat4': 0x8f48,  # GL_DOUBLE_MAT4
```

You can define

```py
class Spv:
    INT32 = 1 << 0
    UINT32 = 1 << 1
    FLOAT32 = 1 << 2
    FLOAT64 = 1 << 3

    VEC2 = 1 << 10
    VEC3 = 1 << 11
    VEC4 = 1 << 12

    MAT2 = 1 << 20
    MAT3 = 1 << 21
    MAT4 = 1 << 22

    UNKNOWN = 1 << 30
```

So

```
'f8 vec2 mat4': 0x8f4d,  # GL_DOUBLE_MAT4x2
```

becomes

```
Spv.FLOAT64 | Spv.VEC2 | Spv.MAT4: 0x8f4d,
```

_moderngl.py
```diff
@@ -366,5 +366,168 @@ def make_varying(name, number, array_length, dimension):
     return res
 
 
+def parse_spv_inputs(raw_spv: bytes):
+    spv = struct.unpack(str(len(raw_spv)//4)+'I', raw_spv)
```
szabolcsdombi, line 370: do not parse everything. make it lazy by using the lambda i previously suggested.

for your example it might be a bit slower, but for actual code where the type information is a small part compared to the computation logic it would matter more.

_moderngl.py
compik710: There have been cases when built-in variables appeared when searching for names. As I recently said gl_InstanceID and the like. 
Well, some of them did not have their typing, and, accordingly, they were not in the allowed (that is, mentioned in SPIR-V) types, so I put such an exception.

_moderngl.py
compik710: OK

_moderngl.py
compik710: Perhaps you're right, I've already tried lambda recently, but I didn't like the speed then. I'll put it on again.

compik710:
Done, the pull request is ready. Here I gave my all. Now it's up to you!
In my opinion, the parser is something new that has never been in ModernGL.

Yeah, in history they will think that I forgot how to count attempts at correction :smile:
But I finally got it! :rocket:

szabolcsdombi:
Thank you, I am going to merge this and add my changes on top of it.

szabolcsdombi:
> Oh, no, it looks like your repository has strong typing... I'm afraid I'll have to redo the whole parser.

not in the core for sure

PR #594, opened 2023-07-16
Improved SPIR-V Program Support

szabolcsdombi:
- [x] Programs accept attribute names as input (necessary for optimized spirv with no debug information)
- [x] VertexArrays accept attribute locations just like they accept attribute names
- [x] Program objects can map attribute locations to gl type information
- [x] Program objects can map attribute names to location for spirv
- [ ] Validate optional attribute names when debug information is present (or glsl is the source)


szabolcsdombi:
@compik710 

Please test this with your use case.

Here are the options:

```py
prog = ctx.program(
    vertex_shader=open('win.vert.spv', 'rb').read(),
    fragment_shader=open('win.frag.spv', 'rb').read(),
    attributes=['in_vert', 'in_color'],  # this is new
)
vao = ctx.vertex_array(prog, buf, 'in_vert', 'in_color')
```

```py
prog = ctx.program(
    vertex_shader=open('win.vert.spv', 'rb').read(),
    fragment_shader=open('win.frag.spv', 'rb').read(),
)
vao = ctx.vertex_array(prog, buf, 0, 1)  # integer locations are new
```

This works too, but it will not be in the release:

```py
prog = ctx.program(
    vertex_shader=open('win.vert.spv', 'rb').read(),
    fragment_shader=open('win.frag.spv', 'rb').read(),
    # no attributes for spm
)
vao = ctx.vertex_array(prog, buf, 'in_vert', 'in_color')
```

The reason the last option won't work is that the names are debug information and may be removed.
Listing the attributes by hand is not a big effort and if anyone worries about failing to maintain this list at two places, the debug names will be used to actually assert on the correctness of this list.

So with debug builds (with names available) the attributes parameter of the program will be validated automatically. (this is not yet implemented)

compik710:
I just checked, everything works as you said. The first option works, the second does not.
```
$ python3 win_test11.py 
  File "/home/user_name/win_test11.py", line 33, in <module>
    vao = ctx.vertex_array(prog, buf, 'in_vert', 'in_color')
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/Github/ModernGL/moderngl/moderngl/__init__.py", line 1908, in vertex_array
    return self.simple_vertex_array(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/Github/ModernGL/moderngl/moderngl/__init__.py", line 1959, in simple_vertex_array
    content = [(buffer, detect_format(program, attributes)) + attributes]
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/Github/ModernGL/moderngl/moderngl/__init__.py", line 2305, in detect_format
    return ' '.join('%d%s' % fmt(types[x] if type(x) is int else types[locations[x]]) for x in attributes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user_name/Github/ModernGL/moderngl/moderngl/__init__.py", line 2305, in <genexpr>
    return ' '.join('%d%s' % fmt(types[x] if type(x) is int else types[locations[x]]) for x in attributes)
                                                                       ~~~~~~~~~^^^
KeyError: 'in_vert'
```

compik710:
Now I'll check on glslc

compik710:
Yes, glslc has the same thing.

compik710:
@szabolcsdombi I found a bug in my parser that concerns names.

szabolcsdombi:
we probably do not handle arrays

```glsl
in vec4 in_data[4]; // from the binding point of view it is the same as mat4 (we parse it as mat4 i think)
in vec4 in_bones[8]; // not sure about this one
```

szabolcsdombi:
not sure if attrib arrays work at all in moderngl. binding these by name does not really make sense as every (sub-) location can be bound separately.

compik710:
I have tested arrays as input, they work in ModernGL. Only I, as I said after writing, "limited myself to the types of variables that were specified in ATTRIBUTE_LOOKUP_TABLE".

Now I will correct my and your work and we will have arrays in SPIR-V.

compik710:
Done, pull request #598
It will be necessary to see if he will pass the builds.

compik710:
So what about the last point?
>- [ ] Validate optional attribute names when debug information is present (or glsl is the source)

Is this a validation here?
```py
ctx.program(attributes=[...])
# ----------------------^^^--
```

szabolcsdombi:
I am planning to merge this as is because it improves/fixes the spv support.
Validation means to check the provided attributes list to be a subset of the detected set.
RIght now the list is not a right choise as a sparse attribute selection is also valid. a dictionary would be a better option.

compik710:
Do you want to do the following and then check for validity?
```py
ctx.program(attributes={location: name, location: name})
```

In my opinion, it sounds somehow too difficult for such a simple task. I think you can make it so that through the attributes option you can set any names, and if one of the locations already has a name, replace it with new ones. If there are more names than attributes detected, cause an error.

And about the names extracted from SPIR-V, I think they should really be removed so that the user does not have problems with switching to an optimized version of shaders. Leave it in the parser, but remove it from ctx.program.

compik710:
It is not worth checking the correctness of the name, the user will enter it twice anyway, or even be able to change it to a more convenient one if the SPIR-V binary is not his.

szabolcsdombi:
for glsl I plan to use `glBindAttribLocation`.

for both glsl and spirv I plan to validate the collected attributes.
This is because on error one might revert to a debug spv build and find the error with a mismatched names.
this is otherwise hard to find/debug.

variables get optimized out and binding not used attributes previusly caused a lot of issues. this might also solve that.

compik710:
Well, since you say there are a lot of problems associated with this, probably your way will be the best.

PR #595, opened 2023-07-18
Improved SPIR-V parsing using no try except

compik710:
### Description

~I haven't found an example of a separate window anywhere, so I'll add my own.~
I decided to abandon the "try:except" construction in the SPIR-V parser, replaced it with a simple check.

### List of changes

- **added** ~example of creating a separate window~
- **fixed** lack of type in the `extracted_general_ids` variable in the SPIR-V parser
- **changed** replaced the "try:except" construction with a check in the SPIR-V parser

### DIFF
--- a/_moderngl.py
+++ b/_moderngl.py
@@ -529,13 +529,12 @@ def parse_spv_inputs(spv: bytes):
         assembly(type_id)
 
     for ids, pointer_type_id in pointer_variable.items():
-        try:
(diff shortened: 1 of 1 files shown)


compik710:
The pull request is ready.

szabolcsdombi:
I like both, and will merge, please split it into 2 different PRs. different changes of any size should have their own PR.

if you are not sure how to do that, remove the window from this one. I will squash merge.

you can make a branch for the window and do the PR from there.

Thank you

PR #596, opened 2023-07-18
Adding an example of creating a separate window

compik710:
### Description

I haven't found an example of a separate window anywhere, so I'll add my own.

### List of changes

- **added** example of creating a separate window


compik710:
Accept my last PR first, then this one. I hope that my commit on the correction of the SPIR-V parser will disappear from here.

PR #598, opened 2023-07-19
Adding support for arrays in the SPIR-V parser

compik710:
### Description

The patch solves the problem using arrays in SPIR-V shaders. At least it passed my tests, including 2-dimensional arrays.

### List of changes

- **added** ability to use arrays as input for SPIR-V shaders


szabolcsdombi:
thank you

PR #604, opened 2023-09-05
Complete addition of DEPTH_CLAMP along with its documentation

compik710:
### Description

Recently faced with the problem of depth clipping when using projection matrices. DEPTH_CLAMP easily solves this problem, so it was not bad to add it here.

### List of changes

- **added** DEPTH_CLAMP and its documentation

### DIFF
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -1472,22 +1472,25 @@ class Context:
         ctx.depth_func = '1'   # GL_ALWAYS
     '''
 
-    depth_clamp: bool
+    depth_clamp_range: Union[Tuple[int, int], None]
     '''
-    bool: Enable/disable depth clamping (``GL_DEPTH_CLAMP``).
-    
-    Depth clamp is needed to disable clipping of fragments outside
+    Setting up depth clamp range (write only, by default None).
+
+    ``ctx.depth_clamp_range`` offers uniform use of GL_DEPTH_CLAMP and glDepthRange.
+
+    ``GL_DEPTH_CLAMP`` is needed to disable clipping of fragments outside
     near limit of projection matrix.
     For example, this will allow you to draw between 0 and 1 in the Z (depth) coordinate,
     even if ``near`` is set to 0.5 in the projection matrix.
     .. Note:: All fragments outside the ``near`` of the projection matrix will have a depth of ``near``.
     See https://www.khronos.org/opengl/wiki/Vertex_Post-Processing#Depth_clamping for more info.
-    
+
+    ``glDepthRange(nearVal, farVal)`` is needed to specify mapping of depth values from normalized device coordinates to window coordinates.
+    See https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDepthRange.xhtml for more info.
+
     Example::
-        # Enable
-        ctx.depth_clamp = True
-        # Disable
-        ctx.depth_clamp = False
+        ctx.depth_clamp_range = None  # It will glDisable(GL_DEPTH_CLAMP) and glDepthRange(0, 1)
+        ctx.depth_clamp_range = (near, far)  # It will glEnable(GL_DEPTH_CLAMP) and glDepthRange(near, far)
     '''
 
     blend_func: Tuple[int, int]
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -151,6 +151,15 @@ BLEND: Constant
 DEPTH_TEST: Constant
 '''Enable/disable depth testing'''
 
+DEPTH_CLAMP: Constant
+'''
+Enable/disable depth clamping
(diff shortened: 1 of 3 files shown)


szabolcsdombi:
This is not a backward compatible change. I think existing user already use the enable_direct for this use case. If we add the depth clamp flag to the enable only, existing unchanged project get a lot of `glDisable(DEPTH_CLAMP)` calls on every `enable_only()` and every scope enter/exit. Probably eliminating the effect of their existing enable_direct calls.

I think it would be possible to add a `ctx.depth_clamp` value.
By default it is None meaning there is no depth clamp range.
When it is set to a (float, float) pair it will call `glEnable(GL_DEPTH_CLAMP)` and `glDepthRange()`.

It is possible we also want a scope variant of this `depth_clamp`.
Maybe it should be called `depth_clamp_range`.

This must be discussed first.
@einarf what do you think?

einarf:
> I think it would be possible to add a `ctx.depth_clamp` value.
> By default it is None meaning there is no depth clamp range.
> When it is set to a (float, float) pair it will call `glEnable(GL_DEPTH_CLAMP)` and `glDepthRange()`.

More fan of this solution 👍

compik710:
The point is that DEPTH_CLAMP takes into account clipping not only for glDepthRange, but also for the perspective matrix used in the vertex shader. So it would be better to make a separate DEPTH_CLAMP control.

szabolcsdombi:
Some related documentations:
[gl core spec 4.5](https://registry.khronos.org/OpenGL/specs/gl/glspec45.core.pdf) page 434
[gl core spec 3.3](https://registry.khronos.org/OpenGL/specs/gl/glspec33.core.pdf) page 100
[glDepthRange](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDepthRange.xhtml)
[glEnable](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glEnable.xhtml)

The spec states the values for clipping are used from the depth range setter function.
So the `ctx.depth_clamp_range: Tuple[float, float] | None` is the proper way to do it in moderngl.

Also it would be a good opportunity to look into `glClipControl(GL_UPPER_LEFT, GL_ZERO_TO_ONE);`
This function call is not always available. It changes the OpenGL coordinate system to be the same as in Vulkan.
I think this should also be added as a boolean property `ctx.clip_control_vulkan` or similar.

compik710:
@szabolcsdombi Looks like I got your idea.

It turns out you need to move DEPTH_CLAMP from `ctx.enable()` to `ctx.depth_clamp`? This will take a little more time, but this way of calling will look quite elegant and solve the backward compatibility problem.
I'll try to do this.

compik710:
The main PR task has been completed.

szabolcsdombi:
depth clamp is just the flag to enable/disable the depth range set by glDepthRange
This can be represented with a single property on the python side `depth_clamp_range`
internally in C there are 3 values, `bool depth_clamp` you already implemented this, and `float depth_range[2]`
I do not consider setting the `depth_clamp` boolean flag from python pythonic enough. You should not think in flags.
The way you can control the flag is by setting the range to a float pair or None.
when setting the range to None, it is equivalent of disabling the range as one would expect.
setting the range to a float pair sets the clamp to enable.
This way the user does not have to maintain two values serving the same purpose.

I review this a bit later when I have more time available. I am sure we will have this feature very soon.

compik710:
I checked `glDepthRange` through PyOpenGL, it simply normalizes the depth for example from [0, 1] to [0.8, 1] and the image does not change. There is no mention of `DEPTH_CLAMP` in the documentation for `glDepthRange`.

https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDepthRange.xhtml

compik710:
If I understand you correctly:
By default `ctx.depth_clamp_range = None`
And if there is a call from the user:
- `ctx.depth_clamp_range = None` It will `glEnable(GL_DEPTH_CLAMP)` and `glDepthRange(0, 1)`
- `ctx.depth_clamp_range = (near, far)` It will `glDisable(GL_DEPTH_CLAMP)` and `glDepthRange(near, far)`

I'm worried about the functionality being cut.
Do you think `glDepthRange(near, far)` is not needed with enabled `GL_DEPTH_CLAMP`?

compik710:
@szabolcsdombi PR is ready.
Replaced `ctx.depth_clamp` with `ctx.depth_clamp_range`.
And also added a few lines to the documentation.

The solution is interesting, but it worked for me.

compik710:
So, do you need anything else for PR?
Everything seems to be ready. For `ctx.clip_control_vulkan` you need another PR.

Although in this case it is worth thinking about ModernVulkan~

szabolcsdombi:
I take a look now,

`ctx.clip_control_vulkan` will be another PR, I probably make it myself.
It can be used for vulkan interoperability or when someone wants to prepare porting the app to vulkan. We already have spv support, we lack this flag.

PR #608, opened 2023-09-26
Fix ctx.depth_clamp_range documentation in __init__.pyi

compik710:
### Description

I didn’t find the addition of `Context.depth_clamp_range` on the documentation site.
It looks like there are problems with readthedocs again, just in case I added some empty lines and at the same time changed the `int` type to `float`. It should work now.

### List of changes

- **fixed** ctx.depth_clamp_range documentation
- **changed** type `int` to type `float`

### DIFF
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -1472,9 +1472,9 @@ class Context:
         ctx.depth_func = '1'   # GL_ALWAYS
     '''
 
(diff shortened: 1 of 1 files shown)


szabolcsdombi:
Indeed, reStructuredText has its own quirks.

compik710:
I still don’t see `ctx.depth_clamp_range` documentation...
It looks like the problems are on the readthedocs side.

Let's wait a week, maybe it will improve.

PR #609, opened 2023-09-26
Fix the visiblity of ctx.depth_clamp_range description in the documentation

compik710:
### Description

Type `Union[Tuple[float, float], None]` python see as `typing.Optional[typing.Tuple[float, float]]`
Now, the type of `ctx.depth_clamp_range` in the docs is just `Tuple[float, float]`.
This is the last way to solve the problem.

### List of changes
- **fixed** visibility of `ctx.depth_clamp_range` in documentation

### DIFF
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -1472,7 +1472,7 @@ class Context:
         ctx.depth_func = '1'   # GL_ALWAYS
     '''
 
-    depth_clamp_range: Union[Tuple[float, float], None]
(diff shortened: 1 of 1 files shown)


szabolcsdombi:
did you also touch any rst files? without that the docs won't change. if you just make blind changes try to build the docs locally with `python setup.py build_sphinx` or look at what the CI is doing, I do not remeber.

compik710:
@szabolcsdombi I completely forgot to update `context.rst`, I already created a PR #610. :sweat:

PR #612, opened 2023-09-28
Tweak Documentation

szabolcsdombi:
This is a draf for the following:

- get rid of autodoc.
- decouple documentation from annotations.
- have the original docs rendered without lost information
- have short but helpful annotations in the stubs

By decoupling the documentation from the annotation, I try to address the following:
Documentation is a long, "example plenty" resource on how and why to use something.
Annotation is a short note on what object are you dealing with.

broad example for Context:

Documentation:

```
The root object exposing OpenGL features.
ModernGL objects can be created from this class.
Contexts are created exclusively from moderngl.create_context,
early in the program's lifecycle, usually right after a window was created.
```

Annotation:

```
root object created by create_context(), it can be used to create buffers, shaders, ...
```


szabolcsdombi:
WIth the annotations I would like to help the users immediately know what the specific object/member/method is for without much useful details. Too much detail on how it works hides the valuable information of "do I need this?" or "do I need something else?"

With the documentation I would like to clarify all details, preferred methods, proper use, ...

compik710:
It turns out you want to make two documentations: for the IDE and readthedocs?
I think I would really like to get a short hint from the IDE right away, but in my opinion everything turns out to be too cumbersome.

I usually write a short description first and then explain in detail, with links. Isn't that how it's usually done?

szabolcsdombi:
You got it right, I think we can also add links to the online docs that vscode for example can open with control click.

I am halfway through the changes necessary to get this working.
I am making bulk moves of text, editing where necessary, also organizing a bit.

When it is done, it should be trivial to edit/extend it. Right now the work cannot be split as it is just tedious copy+edit+check.

I hope I finish it soon.

szabolcsdombi:
Most of the work is complete. All code is on master and it is possible to create PRs from now on.

I do not have time to allocate right now, but I try to find some to fix some bugs I caused.
If you want to poke the docs related to your changes, or any other docs, please feel to do so.

PR #625, opened 2023-10-19
Update examples/creating_separate_window.py

compik710:
### Description

The code didn't work correctly due to `gl_FragDepth = distance;` (fragment shader) on Nvidia video cards, so I replaced it with a projection matrix.
On Nvidia video cards, the previous code showed incorrect display of triangles (that is, with the depth fighting effect). The new code aims to fix this problem.

### List of changes

- **added** projection matrix
- **removed** `gl_FragDepth = distance;` in fragment shader
- **fixed** code operation on Nvidia video cards


compik710:
Hello! This PR is for fixing separate window example code (which was previously created by me).
No further changes are planned. PR is ready.

szabolcsdombi:
thank you

PR #627, opened 2023-11-12
Improve consistency within PyOpenGL example.

whydoubt:
### Description

In the main README.md, there is a small example of PyOpenGL usage for comparison.  There was some inconsistency in how methods were called.  This is a trivial change to make the usage more consistent.


einarf:
Should probably add `GL.` prefix for the enum values as well? That's how most people use pyopengl as far as I have seen.

szabolcsdombi:
I will add those changes. thank you.

PR #638, opened 2023-12-12
Adding direct begin()/end() Query methods

compik710:
### Description

Using the "with" construct to enclose part of the code is inappropriate in some situations. Sometimes you just need to measure the number of rendered samples and because of this, shifting the entire code by 4 spaces can be inconvenient.

Therefore, I suggest adding the following methods. I'll add an example in the documentation later.

How adding will fix this situation:
```py
# Old
while 1:
    with query:
        # ... large rendering code ...
        # because of the "with" you will have to indent

    print("result:", query.samples)
```
```py
# New
while 1:
    query.begin()
    # ... large rendering code ...
    query.end()

    print("result:", query.samples)
```

### List of changes

- **added** direct begin()/end() Query methods

### DIFF
--- a/moderngl/__init__.py
+++ b/moderngl/__init__.py
@@ -117,6 +117,12 @@ class Query:
     def __exit__(self, *args):
         self.mglo.end()
 
+    def begin(self):
+        self.mglo.begin()
+
+    def end(self):
+        self.mglo.end()
+
     @property
     def samples(self):
         return self.mglo.samples


compik710:
For the `ConditionalRender` and `Scope` classes I can also make this little addition.

I look forward to your opinion on the changes.

szabolcsdombi:
this must be used through a context manager

```py
with query.crender:
  ...
```

indent is ok


if you really really do not want the indent, just use the dunder methods directly.

```py
query.crender.__enter__()  # this is working but advised against
...
query.crender.__exit__()
```

compik710:
The `with` construct is indeed very suitable for small code, but for a large project such restrictions can be extremely inconvenient.

Let's imagine the following situation:
```py
cam = engine.camera(count_rendered_samples=True)

while 1:
    cam.render()
    print('FPS:', cam.get_fps(), 'Samples:', cam.get_number_of_rendered_samples())
```

Then in the **camera** class you will have to do something like this:
```py
# Camera class
...
def render(self):
    if self.count_rendered_samples:
            with self.query:
                self._render()  # calling the inner class to avoid large code duplication here
    else:
        self._render()

    self._update_fps_counter()

def _render(self):
    # ... large rendering code ...
...
```
The code does not look very reliable, and if you add, for example, **Scope**, you will have to create a `_render2()` method with similar awkward attachments.
<br>
But if we add the proposed functions, the code can be reduced to the following:
```py
# Camera class
...
def render(self):
    if self.count_rendered_samples:
        self.query.begin()

    # ... large rendering code ...

    if self.count_rendered_samples:
        self.query.end()
    
    self._update_fps_counter()
...
```

I see only advantages here. The code can be written, as you said, with `__enter__` and `__exit__`, but in practice this is a call to internal methods, which is not at all elegant.

szabolcsdombi:
alternatively you can wrap Query in your code.

```py
class MyQuery:
  def __init__(self, query):
    self.query = query
  def begin(self):
    self.query.__enter__()
  def end(self):
    self.query.__exit__()
```

compik710:
But... you understand that you simply moved the problem to another part of the code?
No one will do it like you. Overriding a variable class is natural for Python, but at the cost of performance. Moreover, `__enter__` and `__exit__` do not disappear from the code. The code just grew pointlessly.

The `__enter__` and `__exit__` methods belong entirely to `with`. They are not even mentioned in the user documentation. No one will use them the way you do. **What you show is a temporary solution.**

It seems to me that you are trying to help me. But I want to eradicate the problem at the root.


Is my proposal somehow at odds with the development vector of ModernGL? Simplification is good. But why limit the user so much?

szabolcsdombi:
actually begin end will be there when I remove the Python layer.
Currently we have an extra layer mostly to support typehints which were improved on the python side rendering our layer useless.

This would happen because every object will be replaced with their `.mglo` attribute.
and the mglo has those functions.

Having this method is not a problem unless we document it, which I do not want.
It is fine if you or anyone else would use it, I do not want to document it to prevent confusion around how the Query object works.

Actually I am more reluctant on the conditional render and scope.
The query itself is the most acceptable to have the begin end methods.

Your problem is actually different from the main use case of the query which was for conditional rendering.

```py
with query:
  render_hull()
  
with query.crender:
  render_high_lod()
```

What you want the query for is to probably collect per frame stats.

Soon I will try to backport [`new_frame()` and `end_frame()`](https://zengl.readthedocs.io/en/latest/#Context.new_frame) from zengl.
For moderngl it can have more than just the frame_time query.

compik710:
Now it’s clear that in this case the changes to `moderngl/__init__.py` will be truly meaningless.

And yet, where does `Buffer.orphan()` go?

compik710:
Okay, let this be a surprise. 🙂️

szabolcsdombi:
Orphan will work. I just remove the doca for it. Call it deprecated and that is it.
Orphan prevents us from using immutable storage when available. It is also blocked by the different buffer create flags so orphan is not the only blocker.

PR #642, opened 2024-01-01
Copyright date update in documentation (2024)

compik710:
### Description

Modern documentation with modern copyrights.

### List of changes

- **changed** copyright date in documentation from 2023 to 2024


compik710:
According to UTC time it is already 2024. That's why I'm merging right now.

Issue #670, opened 2024-02-20
Data size mismatch error when creating large textures/size overflow?


jhoadleyCE:
I'm encountering a strange issue when trying to create a texture with a very large size. Specifically, I get a data size mismatch error where both reported sizes are equal and negative. Here's a minimal example to reproduce the issue:

```
import numpy as np
import moderngl

ctx = moderngl.create_context(standalone=True)

width = 14176
height = 10640
channels = 4

input_img = np.zeros((height, width, channels), dtype=np.float32)
texture = ctx.texture((width, height), channels, dtype='f4')  # Fine if I don't try to load data
texture.write(input_img.tobytes())  # Error

# Alternatively
# texture = ctx.texture((width, height), channels, input_img.tobytes(), dtype='f4')  # Error
```

The error is as follows:
Error: data size mismatch -1881645056 != -1881645056

**relevant information:**
moderngl version: 5.10.0
numpy version: 1.25.2
python version: 3.9.18
operating system: windows 11 Business 10.0.22621 N/A Build 22621

compik710:
@jhoadleyCE The problem is solved, you can install the latest version of ModernGL from this repository (instructions in README.md):
```
git clone git@github.com:moderngl/moderngl.git
cd moderngl
python -m pip install -e .
```

compik710:
In some parts of the [code](https://github.com/moderngl/moderngl/blob/main/src/moderngl.cpp), there are still variables of type `Int(32)` that store the sizes of the buffers. It is necessary to get rid of them by retyping from `int` to `unsigned long long` or `Py_ssize_t` depending on parts of the code.

PR #671, opened 2024-02-20
Reliability of calculating expected sizes of buffers, textures, etc.

compik710:
The calculation of the sizes of textures, buffers and the like is now carried out in a variable of the `Int(32)` type. To provide support for larger textures, as well as overall reliability, I change the `expected_size` variable type to `UnsignedLongLong(64)`.

In the future, it will also be necessary to replace other `Int(32)` variables to `UnsignedLongLong(64)` (in particular for buffers), but for now I’m just improving the minimal reliability.

Issue: #670

- **changed** all `int expected_size` to `unsinged long long expected_size`

### DIFF
--- a/src/moderngl.cpp
+++ b/src/moderngl.cpp
@@ -1865,7 +1865,7 @@ static PyObject * MGLFramebuffer_read_into(MGLFramebuffer * self, PyObject * arg
         read_depth = true;
     }
 
-    int expected_size = viewport_rect.width * components * data_type->size;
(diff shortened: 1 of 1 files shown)


compik710:
No problems or bugs were found, I am merging. https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBufferData.xhtml talks about full support for all types of integer variables for `size`.

szabolcsdombi:
not all compiler support `ull = l`. instead also cast the lhs of the operation

```c++
unsigned long long expected_size = (unsigned long long)width * ...
```

PR #680, opened 2024-03-23
Add a pyinstaller hook for easier distribution

Starbuck5:
When moderngl is built into a pyinstaller executable with default settings, it fails at runtime with this error `ModuleNotFoundError: No module named 'glcontext'`

This is because Pyinstaller can only follow imports in Python, and glcontext is imported in C++ code.

To get around this, this commit adds a pyinstaller hook that adds glcontext as a hidden import, tipping pyinstaller off to add that to the build.

Users can achieve this same outcome with Pyinstaller's `--hidden-import` flag, but this will do it automatically so users don't need to worry about it.

Documentation: https://pyinstaller.org/en/stable/hooks.html
We do the same thing in pygame-ce. https://github.com/pygame-community/pygame-ce/tree/main/src_py/__pyinstaller

User reports of this problem:
https://stackoverflow.com/questions/72822188/freezing-python-moderngl-application-with-pyinstaller
https://discord.com/channels/772505616680878080/1220826955688054844

### DIFF
--- a/moderngl/__pyinstaller/__init__.py
+++ b/moderngl/__pyinstaller/__init__.py
@@ -1,0 +1,5 @@
+import os
+
+
+def get_hook_dirs():
+    return [os.path.dirname(__file__)]
--- a/moderngl/__pyinstaller/hook-moderngl.py
+++ b/moderngl/__pyinstaller/hook-moderngl.py
@@ -1,0 +1,1 @@
+hiddenimports = ["glcontext"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[options.entry_points]
+pyinstaller40 =
+	hook-dirs = moderngl.__pyinstaller:get_hook_dirs
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ setup(
     keywords=keywords,
     include_package_data=True,
     package_data={"moderngl-stubs": ["__init__.pyi"]},
-    packages=["moderngl", "moderngl-stubs"],
+    packages=["moderngl", "moderngl-stubs", "moderngl.__pyinstaller"],
     py_modules=["_moderngl"],
     ext_modules=[mgl],
     platforms=["any"],


compik710:
I've always compiled ModernGL like this:
```py
pyinstaller --collect-all moderngl path-to-file.py
```
We planned to make `glcontext` an optional dependency, so I have my doubts.

Starbuck5:
I've never seen anyone use the `--collect-all` option before. I tested, and like `--hidden-import` it successfully picks up on glcontext.

However it took a long time to pick that up:
```
1500 INFO: Determining a mapping of distributions to packages...
99897 INFO: Packages required by moderngl:
['glcontext']
```

`--hidden-import` makes a faster build, and this PR makes it happen automatically without the user needing to do something special for moderngl.

>  We planned to make glcontext an optional dependency, so I have my doubts.

Even an optional dependency should be represented to PyInstaller, because things should work by default. If glcontext isn't present in the environment it will just proceed without it.

compik710:
Then everything is fine, thanks for the PR.

The PR will not affect the core operation of ModernGL and will make it easier for new users to compile executables. Therefore, I decide to merge.

szabolcsdombi:
Let's try:

```py
try:
    import glcontext
except:
    pass
```

in the __init__.py and remove the installer specific name and package

compik710:
I tried both ways, however neither of them worked.

> Let's try:
> 
> ```python
> try:
>     import glcontext
> except:
>     pass
> ```
```py
$ pip install git+https://github.com/moderngl/moderngl.git@<commit hash>
$ pyinstaller --onefile main.py
$ ./dist/main
```
```py
  File "creating_separate_window.py", line 3, in <module>
    ctx = gl.create_context(standalone=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "moderngl/__init__.py", line 1950, in create_context
ModuleNotFoundError: No module named 'glcontext'
```

I built ModernGL via GitHub, main branch, and also tried your suggestion in my fork, but the error remained the same, as if nothing had changed. :thinking:

@Starbuck5, how did you check that PyInstaller is working?

szabolcsdombi:
please check if the collected moderngl has the code I pasted.
I have no idea how pyinstaller collects dependencies.
I am not using pyinstaller because Python supports an official embedded zip for my target platform.

Also please check if after import moderngl, 'glcontext' in sys.modules is present (when glcontext is installed)

szabolcsdombi:
the _pyinstaller package can stay until we find a fix, however with the new loader, rendering to a window requires no glcontext install. we should also test that.

standalone context is not a good test, why would someone use both pyinstaller and a headless context.

Starbuck5:
> @Starbuck5, how did you check that PyInstaller is working?

My test case was extracted from conways game of life from the moderngl examples. My command was `pyinstaller conway.py --hidden-import moderngl_window.context.pyglet`. (Yes, I know, still a hidden import, but I'm trying to fix moderngl for this, not moderngl window).

@szabolcsdombi I think you're right, a hidden import section of python code would work well and be more agnostic. I had thought of doing that, but I wasn't thinking about other packaging solutions. I did it this way because I figured if you ever needed to do more complex things with pyinstaller hooks in the future (like in pygame-ce we manage default assets, binary bundling), it would be good to have the infrastructure in place.

In pygame-ce, we have this code for that purpose: https://github.com/pygame-community/pygame-ce/blob/main/src_py/__init__.py#L343-L349. The imports don't need to actually happen for it to be expressed to these tools, that function is not called.

One confusing bit of this I've run into while looking at this again today is that the conway's game of life example does require glcontext in moderngl's latest release, but not on moderngl built from main?

szabolcsdombi:
https://github.com/pygame-community/pygame-ce/blob/main/src_py/__init__.py#L343-L349

is a very nice solution, we can have it

Starbuck5:
I can make a quick PR for it, but I'm less confident in my testing strategy than I was yesterday.

szabolcsdombi:
> One confusing bit of this I've run into while looking at this again today is that the conway's game of life example does require glcontext in moderngl's latest release, but not on moderngl built from main?

For windowed rendering I am making moderngl standalone and does not require glcontext.
The latest release can use a new method which already does not require glcontext.
On master I already changed the old method to skip glcontext in the most used use-case.

https://github.com/moderngl/moderngl/commit/dc448ed4a76eb1998e558f9e914622545d772c3d

Not yet released, because I had no time to document it. Plus we have some bug on master related to color mask which I also had no time to dig deeper in.

On the latest release try to use `moderngl.get_context()` and it should require no glcontext.
Also this method can be called any number of times.
to explicitly initialize the context createion use init_context()

sorry no docs yet

szabolcsdombi:
You raised this issue, and adding a function to the __init__.py can not really break anything.
I will have to test master before release anyways as there is some breaking code for the color mask

PR #681, opened 2024-03-23
Add hidden imports to init.py

Starbuck5:
Alternate solution for #680, reverts the changes of #680.

Makes sure glcontext can be picked up as a dependency by code-freezing tools, so that programs can be easily packaged into executables without any special actions by users.

Reviewers: if there's anything you'd like to change with this, feel free to push to my branch.

### DIFF
--- a/moderngl/__init__.py
+++ b/moderngl/__init__.py
@@ -12,6 +12,12 @@ except ImportError:
 __version__ = "6.0.0"
 
 
+def packager_imports():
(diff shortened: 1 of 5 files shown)


szabolcsdombi:
Thank you, this looks much simpler. Just for context, I liked the previous solution less because it had to change multiple files of what moderngl contains, and it would have clearly broke for example my web builds of moderngl, where I have to manually copy in each folder in a Containerfile to build it with pyodide-build.
Thank you!
merging

PR #689, opened 2024-04-13
Add support for Per-Fragment Logical Operations (COLOR_LOGIC_OP) (Section 17.3.9 in 4.6 core)

hrbigelow:
Hi,

I wondered if you might consider this, it is added support for OpenGL's target `COLOR_LOGIC_OP` for `gl.Enable`, which allows 'blending' using bitwise logical ops.

I basically just followed the style and conventions in `moderngl/__init__.py` and `src/moderngl.cpp`.

Also added a test script in examples/color_logic_ops.py

Synopsis:

```python
import moderngl as gl
ctx.enable(gl.COLOR_LOGIC_OP)
ctx.logic_op = gl.OR 
# logic_op can be any bitwise op as listed in Table 17.3
# CLEAR, AND, AND_REVERSE, COPY, AND_INVERTED, NOOP, XOR
# OR, NOR, EQUIV, INVERT, OR_REVERSE, COPY_INVERTED, OR_INVERTED
# NAND, SET
...
```

See [4.6 Core Specification](https://registry.khronos.org/OpenGL/specs/gl/glspec46.core.pdf)
image

### DIFF
--- a/examples/color_logic_ops.py
+++ b/examples/color_logic_ops.py
@@ -1,0 +1,96 @@
+'''
+Demonstrates Per-Fragment Blending operation called Logical Operation.
+Instead of blending fragment color based on linear combinations with alpha
+channel, uses bitwise logical operations.
+
+Defined in Section 17.3.9 of
+https://registry.khronos.org/OpenGL/specs/gl/glspec46.core.pdf
+
+Renders four overlapping squares.
+Square i [0, 4) is assigned a red channel with value (1<<i).
+
+'''
+import numpy as np
+import moderngl as gl
+from _example import Example
+
+class LogicalOp(Example):
+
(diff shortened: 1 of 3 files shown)


hrbigelow:
Here is the example generated from `examples/color_logic_ops.py`.  In it, four partially overlapping squares are drawn.  Each square is labeled with a bitmask 1, 2, 4, or 8.  When they overlap they are combined with `gl.OR` creating values between 1 and 15.  This value is then scaled by 1/16 and used for red and blue channels to make variations of purple.

image

szabolcsdombi:
sorry for long time no response, looks good to me.
this will need to wait for a major version update, users may use color op with pyopengl and moderngl.enable may work differently for them as previously this flag was not maintained by moderngl.

PR #699, opened 2024-07-31
Adding automatic compilation of GLSL code to SPIR-V

compik710:
Hello, I propose adding automatic compilation of GLSL shader code to SPIR-V. The new method uses `glslangValidator` call via `subprocess` and thus precompiles the code in SPIR-V, and also writes more detailed errors (in most cases). The method is built into `Context.program()` and `Contex.compute_shader()`, its call is provided by enabling the `to_spirv` flag, which is disabled by default.

At the moment, only `mgl.ComputeShader` work well. There is a bug in `mgl.Program` related to a problem extracting `varyings` from a shader and I need help figuring it out.

<br>
<details>

<summary>Example of use (ComputeShader)</summary>

```py
import moderngl as mgl
import numpy as np


ctx = mgl.create_context(standalone=True)

comp = ctx.compute_shader("""
#version 430

layout(local_size_x = 1024, local_size_y = 1, local_size_z = 1) in;

layout(std430, binding = 0) buffer InputA {
    float valuesA[];
};

layout(std430, binding = 1) buffer InputB {
    float valuesB[];
};

layout(std430, binding = 2) buffer Results {
    float results[];
};

layout(std140, binding = 3) uniform Result_size {
    uint results_size;
};

void main() {
    uint idx = gl_GlobalInvocationID.x;
    if (idx < results_size) {
        results[idx] = valuesA[idx] * valuesB[idx];
    }
}
""", to_spirv=True)

valuesA = np.array([2, 4, 6], dtype=np.float32)
valuesB = np.array([3, 5, 7], dtype=np.float32)

bufferA = ctx.buffer(valuesA)
bufferB = ctx.buffer(valuesB)
result_buffer = ctx.buffer(reserve=bufferA.size)

res_size_data = np.array([len(valuesA)], dtype=np.uint32)
res_size_buffer = ctx.buffer(res_size_data)

bufferA.bind_to_storage_buffer(0)
bufferB.bind_to_storage_buffer(1)
result_buffer.bind_to_storage_buffer(2)
res_size_buffer.bind_to_uniform_block(3)

comp.run(valuesA.shape[0] // 1024 + int(valuesA.shape[0] % 1024 > 0))

result_data = np.frombuffer(result_buffer.read(), dtype=np.float32)
print(result_data)
```
```sh
$ python3 t3.py 
[ 6. 20. 42.]
```
</details>

<details>

<summary>Example of use (Program (transforms)  (not works for now))</summary>

```py
import moderngl as mgl
import numpy as np


ctx = mgl.create_context(standalone=True)

prog = ctx.program(vertex_shader="""
#version 430

layout(location = 0) in float var1;
layout(location = 1) in float var2;

layout(location = 0) out float var3;

void main() {
    var3 = var1 * var2;
}
""", varyings=('var3',), to_spirv=True)

values = np.array([
    2, 3,
    4, 5,
    6, 7
], dtype=np.float32)

vbo = ctx.buffer(values.tobytes())
vao = ctx.vertex_array(prog, [
    (vbo, 'f f', 'var1', 'var2')
])

num_vertices = len(values) // 2

result_buffer = ctx.buffer(reserve=num_vertices * 4)

ctx.enable(ctx.RASTERIZER_DISCARD)
vao.transform(result_buffer, vertices=num_vertices)
ctx.disable(ctx.RASTERIZER_DISCARD)
ctx.finish()

result_data = np.frombuffer(result_buffer.read(), dtype=np.float32)
print(result_data)
```
```sh
$ python3 t3.py 
  File "/home/user/t3.py", line 36, in <module>
    vao.transform(result_buffer, vertices=num_vertices)
  File "/home/user/Github/ModernGL/moderngl/moderngl/__init__.py", line 1181, in transform
    self.mglo.transform(outputs, mode, vertices, first, instances, buffer_offset)
_moderngl.Error: the program has no varyings
```
</details>

### DIFF
--- a/_moderngl.py
+++ b/_moderngl.py
@@ -440,31 +440,36 @@ def make_varying(name, number, array_length, dimension):
 
 
 def convert_glsl_to_spirv(glsl_code, shader_type):
-    if shutil.which('glslangValidator') is None:
-        raise RuntimeError("glslangValidator not found. Make sure it is installed and available in PATH")
-    
+    supported_utils = ['glslangValidator', 'glslc']
+    available_utils = [util for util in supported_utils if shutil.which(util)]
+
+    if len(available_utils) < 1:
+        raise RuntimeError("glslangValidator or glslc were not found. Make sure it is installed and available in PATH")
+
     if isinstance(glsl_code, bytes) and int.from_bytes(glsl_code[:4], "little") == 0x07230203:
         raise ValueError("The code has already been converted to SPIR-V, it is impossible to convert it again")
-    
+
     if shader_type not in ['vert', 'tesc', 'tese', 'geom', 'frag', 'comp']:
         raise ValueError(
             f"Unknown shader type. The following are available: vert, tesc, tese, geom, frag, or comp. Not {shader_type}")
-    
-    if isinstance(glsl_code, str):    
+
+    if isinstance(glsl_code, str):
         glsl_bytes = glsl_code[glsl_code.find('#'):].encode('utf-8')
     else:
         glsl_bytes = glsl_code[glsl_code.find(b'#'):]
 
-    current_platform = platform.system().lower()
-    if current_platform == 'windows':
+    output_file = '/dev/stdout'
+    if platform.system().lower() == 'windows':
         output_file = 'CON'
-    elif current_platform in ['linux', 'darwin']:
-        output_file = '/dev/stdout'
-    else:
-        raise OSError(f"OS {platform.system()} not supported")
+
+    command = []
+    if available_utils[0] == supported_utils[0]:
+        command = ['glslangValidator', '--stdin', '-S', shader_type, '-V', '-o', output_file]
+    elif available_utils[0] == supported_utils[1]:
+        command = ['glslc', f'-fshader-stage={shader_type}', '-', '-o', '-']
 
     process = subprocess.Popen(
-        ['glslangValidator', '--stdin', '-S', shader_type, '-V', '-o', output_file],
+        command,
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE
--- a/_moderngl.py
+++ b/_moderngl.py
@@ -1,4 +1,7 @@
 import struct
+import shutil
+import platform
+import subprocess
 from typing import Any, Dict, List, Tuple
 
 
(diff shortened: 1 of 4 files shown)


compik710:
Looks like #592, later I will write another parser and fix the problem. Basically, the PR is ready, so I'm starting to merge.

compik710:
Hi, we don't really need to create a separate module. I just use the system call and work through `stdin` and `stdout,` so there is no need to allocate it to a separate module. Moreover, dependencies are installed separately, as is done with glcontext. In fact, this is like creating a context in **ModernGL**, although it is done through **glcontext**.

More tests of this approach will need to be conducted, but in fact `glslangValivator` it is not outdated, and `glslc` is just its more convenient shell for use in the console.

I still need to study this issue, but from the first links of the discussion of this point on the Internet, I see that `glslangValidator` compiles easier and has a minimum of dependencies, unlike `shaderc`.

compik710:
I checked the possibility of replacing it with `glslc`, everything works. It remains to figure out exactly how to make a choice between `glslangValidator` and `glslc`.

I suggest making an automatic detection depending on which utilities are available to the user. However, now we need to decide which library will be preferable if both are available... :thinking:

Given that `glslc` is a shell of `glslangValidator`, I am inclined to conclude that for a simple compilation from **GLSL** to **SPIR-V** for **OpenGL**, the `glslangValidator` functionality will be enough and save resources. What is your point of view, @szabolcsdombi?

compik710:
`glslc` support added. Ready to merge.

compik710:
@szabolcsdombi, I have added `glslc` support and updated the documentation. Now the program automatically detects the available GLSL compilation tool in SPIR-V, giving priority to `glslangValidator` if it is available. This provides flexibility and simplifies the process for users without requiring them to make a clear choice.

Regarding a separate module, given that the functionality is unlikely to grow, I think that integration into the current structure is the best way out. It preserves the simplicity and unity of the project.

If you have additional comments or suggestions on the current approach, I am ready to consider them and make the necessary adjustments. I am waiting for your response for further action.

szabolcsdombi:
I have managed to read all the above.

First of all, we have all code that is required to support SPV.
If you want to pass in glsl but compile it to SPV you need to do something like

```py
ctx.program(
    vertex_shader=compile_glsl_to_spv('''
    ''', **compiler_options_here),
)
```

for example this should work out of the box with `from glnext_compiler import glsl`

This does not really makes sense when using OpenGL because GLSL works out of the box.
given the context type (standard, no_error, debug, robust, ...) the glsl compiler in the driver may produce different binaries to better fit the user's needs.

SPV was created to allow the developer to "compile" the shader before the program starts. This does not happen if the program itself compiles the shader. Making SPV useless.

The reason I build glnext_compiler, is that in order to prototype without any tooling installed, one may give up the benefits of SPV and do the compilation on the spot. This is necessary in Vulkan, because there is no GLSL support. So glnext_compiler is a tradeoff.

Detecting glslangValidator or glslc is not a good idea. I think the SDK installs them both anyways.
The SDK is huge. However if you want the compiler glslc itself provides a library file which can be linked into the extension. (glnext_compiler does this). This would allow the user to use glslc without it being installed.

Adding this code to this repo makes no sense, because we don't want to encourage users to use spv in a way that it was not supposed to be used. Also providing a proper compiler would require supporting all the flags, which is definitely not code moderngl is willing to maintain.

If you want to implement this feature I highly encourage you to create a new extension. This is why I also suggested to look into glnext_compiler, basically that statically links glslc and provides a basic compilation already. So you can probably take valuable code from that repo and reuse it in yours.

With a custom extension you should be able to create builder functions for both program and compute shader, mocking the moderngl api. If these functions return moderngl objects. all the other part of moderngl would work. Here is an example:

```py
# let's assume this module is called spv

import moderngl

def compile_glsl_to_spv(...):
    ...

def program(vertex_shader, fragment_shader, ...) -> modenrgl.Program:
    ctx = moderngl.get_context()
    return ctx.program(
        vertex_shader=compile_glsl_to_spv(vertex_shader),
        ...
```

with this module at hand you can:

```py
ctx = moderngl.create_context()

program = spv.program(...)
ctx.vertex_array(program, ...)
```

PR #701, opened 2024-09-05
ADD ModernGL Water Example

rede97:
The **[WebGL Water](https://madebyevan.com/webgl-water/)** example was rewritten use moderngl.
Snipaste_2024-09-05_20-29-28


szabolcsdombi:
Hello!
Thank you for submitting this.
I took a quick look and it seems the original example has no license attached to it.

https://github.com/evanw/webgl-water

There is some extra credited work on the page:

"Tile texture from [zooboing](https://www.flickr.com/photos/zooboing/3682834083/) on Flickr"

We should have this information somewhere. also the number `16_` can be omitted.
The resources are small. I am fine merging this as-is.

Just confirm if the PR is ready and I merge it right away.

PR #702, opened 2024-09-08
Grammar fix in documentation

PycraftDeveloper:
Fixed line: "The maximum number of vertices that.

the geometry shader will output."

To: "The maximum number of vertices that the geometry shader will output."


szabolcsdombi:
thank you!

PR #706, opened 2024-09-30
Add support for debug labels

JesseTG:
This PR adds support for object labels for most objects exposed by moderngl. Here's how that works:

- If using OpenGL 4.3 or greater, standard `glObjectLabel` will be used.
- Otherwise if `GL_EXT_debug_label` is available, `glLabelObjectEXT` will be used.
- Otherwise no OpenGL label functionality will be used, but a `str` will be kept in the Python objects so that client code will work anyway.
- I wanted to include `GL_KHR_debug` in the fallbacks, but I was advised that this is likely unnecessary. So I just left a note in case the issue ever comes up again.

The only applicable object I didn't add label support for is `moderngl.Query`; since it's actually comprised of four [query objects](https://www.khronos.org/opengl/wiki/Query_Object), I couldn't decide on a suitable design. I was thinking about either exposing the labels individually or using one as a common prefix.

### DIFF
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -2825,14 +2825,6 @@ class Query:
     extra: Any
     """Attribute for storing user defined objects"""
 
-    label: str | None
-    """
-    A human-readable name for this object,
-    intended for debugging tools.
-    Will be set or fetched with OpenGL label calls if available,
-    or kept within the Python object if not.
-    """
-
     def __enter__(self): ...
     def __exit__(self, *args: Tuple[Any]): ...
 
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -436,6 +436,14 @@ class Buffer:
     This values is provided for debug purposes only.
     """
 
+    label: str | None
+    """
+    A human-readable name for this object,
+    intended for debugging tools.
+    Will be set or fetched with OpenGL label calls if available,
+    or kept within the Python object if not.
+    """
+
     def write(self, data: Any, offset: int = 0) -> None:
         """
         Write the content.
@@ -651,6 +659,14 @@ class ComputeShader:
     The internal OpenGL object.
 
     This values is provided for debug purposes only.
+    """
+
+    label: str | None
+    """
+    A human-readable name for this object,
+    intended for debugging tools.
+    Will be set or fetched with OpenGL label calls if available,
+    or kept within the Python object if not.
     """
 
     def __getitem__(self, key: str) -> Union[Uniform, UniformBlock, StorageBlock]:
@@ -2497,6 +2513,14 @@ class Framebuffer:
     This values is provided for debug purposes only.
     """
 
+    label: str | None
+    """
+    A human-readable name for this object,
+    intended for debugging tools.
+    Will be set or fetched with OpenGL label calls if available,
+    or kept within the Python object if not.
+    """
+
     def clear(
         self,
         red: float = 0.0,
@@ -2737,6 +2761,15 @@ class Program:
     This values is provided for debug purposes only.
     """
 
+    label: str | None
+    """
+    A human-readable name for this object,
+    intended for debugging tools.
+    Will be set or fetched with OpenGL label calls if available,
+    or kept within the Python object if not.
+    """
+
+
     def get(self, key: str, default: Any) -> Union[Uniform, UniformBlock, Attribute, Varying]:
         """
         Returns a Uniform, UniformBlock, Attribute or Varying.
@@ -2792,6 +2825,14 @@ class Query:
     extra: Any
     """Attribute for storing user defined objects"""
 
+    label: str | None
+    """
+    A human-readable name for this object,
+    intended for debugging tools.
+    Will be set or fetched with OpenGL label calls if available,
+    or kept within the Python object if not.
+    """
+
     def __enter__(self): ...
     def __exit__(self, *args: Tuple[Any]): ...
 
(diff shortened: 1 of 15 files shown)


szabolcsdombi:
Hey! Already very nice, here are some major changes I think would be nice:

- instead of really getting and setting the value, we should store the str inside the python layer `__init__.py` in some `_label` field. Implement a property getter setter for this field called `label` and inside the setter call a magic method `ctx._set_label(type, glo, label)`.

you can implement this Context method exactly as `static int set_label(MGLContext * ctx, GLenum type, GLuint object, PyObject * value) {`, please parse arguments with `PyArg_Parse...`
inside the setter please make label errors ignored.

I prefer if the actual `_set_label` is silent, or returns a True/False based on success.
The label setting in no way should raise an error.
We want to ignore this return value, it's ok to have it there for debug purposes.

JesseTG:
I wanted to delegate to actual API calls (including for retrieving the label later) to handle the case where something else changes the label behind our backs; if the label string in Python says something different than the value returned by `glGetObjectLabel`, that will be a pain in the ass to notice and debug. (Remember, my use case for `moderngl` involves running OpenGL-dependent native code that may already be buggy.) What do you think?

szabolcsdombi:
then add a `ctx._get___` method as well. we don't do this with many other values, including blending settings and similar.

JesseTG:
Let me make sure I understand. Are you suggesting that labels should _primarily_ work the way you've [previously described](https://github.com/moderngl/moderngl/pull/706#issuecomment-2384905235), but that there should be an explicit `Context._get_label` method for the occasional edge case?

szabolcsdombi:
I meant that you can have the getter actually call a function and not just return a stored value.
it would be nice to have one or two functions in C and do the rest from the python layer. If it is still ambiguous, just let me know.

JesseTG:
> I meant that you can have the getter actually call a function and not just return a stored value. it would be nice to have one or two functions in C and do the rest from the python layer. If it is still ambiguous, just let me know.

Is something like this what you had in mind?

```python
class Texture:
    @property
    def label(self):
        return self.ctx.mglo.get_label(self.glo, GL_TEXTURE)

    @label.setter
    def label(self, lbl):
        self.ctx.mglo.set_label(self.glo, GL_TEXTURE, lbl)

    @label.deleter
    def label(self):
        self.ctx.mgl.set_label(self.glo, GL_TEXTURE, None)
```

szabolcsdombi:
yes, looks great!

you can just support '' or None in the setter and have no deleter

```py
@label.deleter
    def label(self):
        self.ctx.mgl.set_label(self.glo, GL_TEXTURE, None)
        ```

szabolcsdombi:
will this getter setter work for all your use cases?

JesseTG:
Yes, this would work fine. I'll revise my patch accordingly.

As for a deleter...I figured that adding one would be the Pythonic thing to do. Since OpenGL allows removing a label by passing `NULL`, that means `del something.label` would have sensible semantics. If you really insist, I can skip adding one.

szabolcsdombi:
Default label is none, setting a value to None should remove it (in gl terms) that is why i suggested you need no deleter. I like setting to None over the deleter.
It would also be counterintuitive to see del foo.label then dir(foo) still has label listed

szabolcsdombi:
This PR makes me think if we would like to auto label objects (only is enabled explicitly at context create).
We can do this in a separate pr.
Basically we could label with str(id(mgl_object)) and even get the object back by returningit as an object using the c api. This obviously can segfault, however it won't happen because we control when an actual object will be deallocated.
Or we can just keep amapping from generated label name to instance.
I'll describe this better when we get this one merged

JesseTG:
> Default label is none, setting a value to None should remove it (in gl terms) that is why i suggested you need no deleter. I like setting to None over the deleter. It would also be counterintuitive to see del foo.label then dir(foo) still has label listed

All right, I'll remove the deleters.

> This PR makes me think if we would like to auto label objects (only if enabled explicitly at context create).

I don't think auto-labeling OpenGL objects is a good idea, as doing so provides no information about how the underlying resource is actually used. This would add complexity without enabling clarity.

JesseTG:
All right, I think this PR is ready for your review.

szabolcsdombi:
auto labeling is only useful if you can retrieve the python object, this would allow accessing arbitrary data on those objects. first I have to see how labels and debug messaging work with moderngl.

szabolcsdombi:
Finally found something about using the core or the extension variants:

Do not use

szabolcsdombi:
`GL_KHR_debug`: https://opengl.gpuinfo.org/listreports.php?extension=GL_KHR_debug
(Showing 12,859 entries (filtered from 13,687 total entries)

but `GL_KHR_debug` was made core, which means function loading for the non KHR names will work as expected.

JesseTG:
> Finally found something about using the core or the extension variants:
> 
> Do not use `GL_EXT_debug_marker` as a fallback, it is pointless: https://opengl.gpuinfo.org/listreports.php?extension=GL_EXT_debug_marker (Showing 469 entries (filtered from 13,687 total entries))

`GL_EXT_debug_marker` is a separate extension, one that doesn't involve object labels. Did you mean [`GL_EXT_debug_label`](https://opengl.gpuinfo.org/listreports.php?extension=GL_EXT_debug_label)? Because Apple devices _do_ implement `GL_EXT_debug_label`, so this is the only way to make up for their lack of `glObjectLabel` support.

szabolcsdombi:
Okay i am convinced ;) keep things as is. We can merge when you say it is ready.

JesseTG:
> Okay i am convinced ;) keep things as is. We can merge when you say it is ready.

It's ready!

szabolcsdombi:
Thank you!

PR #713, opened 2024-11-20
Basic mesh shader support

MaxLunar:
This PR adds basic support for mesh shaders (NVIDIA extension), without multidraw. Might've broken formatting style

Also added an example with demo shaders (courtesy of @usernameak)


usernameak:
Just in case, I gave permission to use the shaders from my old mesh shader demo (https://github.com/usernameak/meshshadervoxels) here.

szabolcsdombi:
Looks good to me. Thank you for contributing.
We are missing the typehints in the .pyi file. I make a separate issue for that.
Do you need this released anytime soon?

szabolcsdombi:
i've seen the reformats i don't mind. if you have time to pull request the mentioned multidraw and stubs it would be great. if not i may add it later and release only after that. you can install anytime from moderngl's main branch with `pip install https://github.com/moderngl/moderngl/archive/refs/heads/main.zip`

PR #714, opened 2024-11-25
Update __init__.pyi Adding more specific doc to set a binding point on UniformBlock

MrDiver:
I think it might be good to add some more documentation to the comments. I am not quite sure which style you are using for your comments so feel free to change it, but that information was not really clear from the Documentation!

Hope that helps, and i am happy to get linked some resources where i can find which style you are using and how i should contribute to the documentation!

### DIFF
--- a/moderngl-stubs/__init__.pyi
+++ b/moderngl-stubs/__init__.pyi
@@ -299,6 +299,11 @@ class UniformBlock:
     binding: int
     """
     The binding of the uniform block.
+    The binding point can be set by assigning a value to this object retrieved by a program
+    
+    Example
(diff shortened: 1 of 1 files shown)


szabolcsdombi:
thank you please feel free to add text like this where you feel the current documentation is not enough.

MrDiver:
Sure, will do! I think i have a lot of different pain points that i found while working on Manim renderer! So i will make sure to include them all in the documentation so that they are easier to find!

szabolcsdombi:
let me know when you think the PR is ready.

a bit unrelated, but since you are an active developer on the ManimCommunity. does the latest moderngl working well? is there anything moderngl lacks that manim would need or needs a workaround for?

PR #715, opened 2024-11-26
Multidraw, typehint & docs stubs, improved example

MaxLunar:
Rewrote `draw_mesh_tasks_indirect` to use multidraw internally (doesn't quite make sense since you can multidraw with `drawcount=1` which is default and optional), added `draw_mesh_tasks_indirect_count`. Improved example (some animation, technically uses indirect for demonstration purposes). Added stubs for mesh shader related stuff and indirect compute shader call.

### DIFF
--- a/examples/mesh_shaders.py
+++ b/examples/mesh_shaders.py
@@ -1,196 +1,198 @@
+import os
+import math
+
 import moderngl
 import glm
 import numpy as np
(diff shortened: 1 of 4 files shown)


MaxLunar:
I didn't check how does the resulting documentation look though, but hope it's fine.

PR #720, opened 2025-01-06
Add task and mesh shaders to Context.program in __init__.pyi

cyuria:
#713 and #715 didn't update the type stubs for `Context.program()`


szabolcsdombi:
thank you

PR #721, opened 2025-01-08
Update moderngl_window examples with new callback names

cyuria:
https://github.com/moderngl/moderngl-window/pull/203 changed the name of the `render` callback to `on_render` (among others). This change hasn't been reflected in the examples here and so they do not run with the latest moderngl_window package from pypi.

The water example is broken on my machine due to deprecated glsl code, which prevents the shaders from compiling, however the other examples do all work.

I've updated parts of the old examples as well, but many of them are broken with being moved into the `examples/old/` directory and I really can't be bothered to test all of them given that they're old anyway.


einarf:
Thank you. Definitely something we forgot.

PR #723, opened 2025-02-15
Gil release during major blocking operations

jimy-byerley:
Most openGL calls are non-blocking because they simply queue commands on the GPU, only some calls are blocking the calling thread. These are call synchonizing with the GPU (explicitely like `finish`  or implicitely while retreiving data or mapping buffers from the GPU).

The current modernGL implementation holds the GIL during all operations, causing the python interpreter to freeze all its threads when one of them performing such a synchronization. 
This is not a concern in single-thread programs, but may be a major performance issue in multithread programs (with other threads doing IOs or computations for instance)

This PR proposes to release the [GIL](https://docs.python.org/3/c-api/init.html#thread-state-and-the-global-interpreter-lock) during these blocking operations

impacted methods:
- `Context.finish`
- `Buffer.read*`
- `Texture*.read*`

buffer writing methods also perform synchonization since moderngl implement them using [`glMapBufferRange`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml), but discussing with @szabolcsdombi, lets consider they should not block for too long when the user do not write to buffers after rendering them

### DIFF
--- a/docs/reference/buffer.rst
+++ b/docs/reference/buffer.rst
@@ -26,14 +26,15 @@ Methods
     :param bytes data: The data.
     :param int offset: The offset in bytes.
 
-.. py:method:: Buffer.read(size: int = -1, *, offset: int = 0) -> bytes:
+.. py:method:: Buffer.read(size: int = -1, *, offset: int = 0, gil=False) -> bytes:
 
     Read the content.
 
     :param int size: The size in bytes. Value ``-1`` means all.
     :param int offset: The offset in bytes.
+    :param bool gil: Whether to hold the GIL while waiting for the data.
 
-.. py:method:: Buffer.read_into(buffer: Any, size: int = -1, *, offset: int = 0, write_offset: int = 0) -> None:
+.. py:method:: Buffer.read_into(buffer: Any, size: int = -1, *, offset: int = 0, write_offset: int = 0, gil=False) -> None:
 
     Read the content into a buffer.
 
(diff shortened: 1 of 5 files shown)


jimy-byerley:
Here is the pull request !
I'm not sure on which target branch you prefer to merge, so tell me and I will resolve the conflicts

jimy-byerley:
**Note:** I tested this PR against the unit tests present on the repo, as well as my own code using moderngl. I do not have however a program using all the features of moderngl together nor any performance benchmark under the hand.
Do you have other programs to test this code ?

szabolcsdombi:
Hey I will take a closer look, but until then:

No gil parameter is needed, just simply add the thread related macros around the only blocking function.

If the gil is not held it is not safe to access anything like `self->context->default_texture_unit`
So please add the macros around the functions that might block like `gl.GetTexImage`

It's fine to have read and finish be slower by two mutex operations.

jimy-byerley:
Sorry for the delay, There is almost nothing to do to finish, but I was very busy since February ...
Hopefully in 2 weeks I will have more time

tpwrules:
I have run into this before with some drivers simply using functions like `Framebuffer.clear`. I'm not sure how to decide a good set of functions, but some definitely should release the GIL.

szabolcsdombi:
in the projects i used moderngl or zengl releasing and reaquiring the gil was not healty for a stable framerate. this is why i would prefer it optional.

Globefishp:
Hello, is there any progress to have an option to release gil when calling ctx.finish()? I'm also prefer a no gil version because there may be I/O in background thread.

jimy-byerley:
i never got back to it, sorry. but not a big work to do :sweat:

PR #726, opened 2025-03-05
Defaults for inactive/nonexistent uniforms

MaxLunar:
Might be related to #663 and #664, allows assigning to uniforms that arent active or dont exist (technically members). Possibly breaking and quick fix, looking for comments

### DIFF
--- a/examples/mesh_shaders.py
+++ b/examples/mesh_shaders.py
@@ -173,7 +173,7 @@ class Example(mglw.WindowConfig):
         self.camera = self.program["modelViewProjection"]
         self.u_time = self.program["time"]
 
-    def render(self, time: float, frame_time: float):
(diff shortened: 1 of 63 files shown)


yoyonel:
Can be cool (necessary) to have a unit test (at least) who ❌fail before your modification and ✅pass after.

szabolcsdombi:
I'm fine having this, please move UniformDummy out and call it FakeUniform or InvalidUniform
in the getter don't instanciate instead use the same if pattern as in the setter.

szabolcsdombi:
sorry for the late reply. i missed this somehow.

MaxLunar:
thats fine, I'll revisit it later when I get more free time this week. This is more of a codestyle question, and I wasn't sure how to do it in that regard

PR #742, opened 2025-12-15
Support array of bindless textures

icaven:
Currently the library supports single bindless texture handles, but not arrays of bindless texture handles - this PR adds that support.

An example program is added showing how they are used.


klattimer:
I've been testing this branch, and I hit the following error. 

TypeError: argument 3 must be int, not list

klattimer:
Nevermind, I had a stale old version getting loaded.

src/moderngl.cpp
```diff
@@ -77,6 +77,49 @@ struct MGLDataType {
     bool float_type;
 };
 
+// Encapsulates bindless texture handle state
+struct BindlessHandleState {
+    unsigned long long handle;
+    bool obtained;
+    bool resident;
+
+    BindlessHandleState() : handle(0), obtained(false), resident(false) {}
```
szabolcsdombi, line 86: this code will never run see similar parts of the code for initializing with zeros.

src/moderngl.cpp
```diff
@@ -8186,19 +8228,154 @@ static PyObject * MGLContext_write_uniform(MGLContext * self, PyObject * args) {
     Py_RETURN_NONE;
 }
 
-static PyObject * MGLContext_set_uniform_handle(MGLContext * self, PyObject * args) {
-    int program_obj;
+static PyObject * MGLContext_set_uniform_handle(const MGLContext * self, PyObject *args) {
```
szabolcsdombi, line 8231: no const here

Johyysminny (approved):

src/moderngl.cpp
icaven: removed

PR #744, opened 2026-01-05
Add support for additional texture wrapping modes

icaven:
The current texture wrapping modes supported by the package are `repeat` or `clamp to edge`, and the setting is controlled by a set of boolean properties on the objects that have wrapping modes (`Sampler`, `Texture`, `Texture3D`, and `TextureArray`).  

The interface shall be changed to support the full set of wrapping modes, but still maintain moderngl API compatibility.  The wrapping modes are `repeat`, `clamp_to_edge`, `clamp_to_border`, `mirrored_repeat`, and `mirror_clamp_to_edge`.

A new field `wrap` shall be added to the above objects, which may be set by a dictionary of wrapping mode values, keyed by the axes that they apply to.  To maintain compatibility with the convention in ModernGL, the names of the axes shall be "x", "y" and "z", but the synonyms "s", "t", "r" shall also be accepted when setting the values. (The synonyms match the OpenGL convention for the labelling of the axes).

The value for the wrapping mode shall be one of those in the above list of values.

When a change to the wrapping mode is made either directly through the new `wrap` property or through the previous repeat_[xyz] interface, the result shall be consistent: for example, setting the wrap property for the 'x' axis to `repeat` shall set the `repeat_x` property to `True`, and vice-versa.  Setting the `repeat_x` property to `False`, shall set the `wrap['x']` property to `clamp_to_edge`, which is the same as previously.  If the `wrap['x']` property is set to one of the other wrapping modes, then the `repeat_x` property shall also be set to `False`.

The values for all axes may be changed at the same time, by setting the `wrap` property to a dictionary; for example: 

```
texture.wrap = {"x": "mirrored_repeat", "y": "clamp_to_edge"}
```

or individually:

```
texture.wrap["x"] = "mirrored_repeat"
texture.wrap["y"] = "clamp_to_edge"
```

If a wrap property is set using a dictionary, then only the keys that are recognized by the object shall be used.  For example, the `Texture` object only uses the "x" and "y" axes, but if the dictionary contains the "z" key too, that key shall be ignored.  If the dictionary contains both the principal axis name and its synonym (for example, "x" and "s"), the values must be the same, otherwise an exception shall be raised.

### DIFF
--- a/src/moderngl.cpp
+++ b/src/moderngl.cpp
@@ -9557,7 +9557,7 @@ static PyGetSetDef MGLTextureArray_getset[] = {
     {(char *)"repeat_x", (getter)MGLTextureArray_get_repeat_x, (setter)MGLTextureArray_set_repeat_x},
     {(char *)"repeat_y", (getter)MGLTextureArray_get_repeat_y, (setter)MGLTextureArray_set_repeat_y},
     {(char *)"wrap", (getter)MGLTextureArray_get_wrap, (setter)MGLTextureArray_set_wrap},
-   {(char *)"filter", (getter)MGLTextureArray_get_filter, (setter)MGLTextureArray_set_filter},
+    {(char *)"filter", (getter)MGLTextureArray_get_filter, (setter)MGLTextureArray_set_filter},
     {(char *)"swizzle", (getter)MGLTextureArray_get_swizzle, (setter)MGLTextureArray_set_swizzle},
     {(char *)"anisotropy", (getter)MGLTextureArray_get_anisotropy, (setter)MGLTextureArray_set_anisotropy},
     {},
--- a/src/moderngl.cpp
+++ b/src/moderngl.cpp
@@ -1,5 +1,7 @@
 #define PY_SSIZE_T_CLEAN
+#include <functional>
 #include <Python.h>
+#include <string>
 
 #include "gl_methods.hpp"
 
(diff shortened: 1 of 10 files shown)


Tremeschin:
+1, I have a ~8% performance loss in [depthflow](https://github.com/BrokenSource/DepthFlow) by needing to manually [juggle ratios](https://github.com/BrokenSource/ShaderFlow/blob/071fa5c70ff93d15cb8df4e99f9a9699a535a3dc/shaderflow/resources/shaders/include/shaderflow.glsl#L118-L132) for a mirrored_repeat mode, so that the depthmap is continuous when out of bounds and doesn't cause interpolation artifacts on edges.

src/moderngl.cpp
```diff
@@ -1,5 +1,7 @@
 #define PY_SSIZE_T_CLEAN
+#include <functional>
 #include <Python.h>
+#include <string>
```
szabolcsdombi, line 4: moderngl is very close to compile as C and not C++ i would like to avoid c++ headers.

src/moderngl.cpp
```diff
@@ -1,5 +1,7 @@
 #define PY_SSIZE_T_CLEAN
+#include <functional>
```
szabolcsdombi, line 2: can we get rid of this? where is it used?

icaven: I removed the C++ headers, and in general replaced C++ code with C

